PR #11, opened 2015-01-15
Changes to the format and clean up.

joachimmetz:
@destijl for review


joachimmetz:
To do: update pull request.

joachimmetz:
@destijl can you move this review forward I'm find myself fixing issues where Registry values are indicated as Registry keys and no documentation links

test_data/definitions.yaml
```diff
@@ -24,9 +24,11 @@ urls: ['http://support.microsoft.com/kb//214653']
...
 name: CurrentControlSet
 doc: The control set the system is currently using.
-collectors:
-- collector_type: REGISTRY_VALUE
-  args: {path_list: ['HKEY_LOCAL_MACHINE\SYSTEM\Select\Current']}
+sources:
+- type: REGISTRY_VALUE
+  attributes:
+    key_value_pairs:
+      - {key: 'HKEY_LOCAL_MACHINE\SYSTEM\Select', value: 'Current'}
```
destijl: It's confusing to have both registry_key and registry_value (that also takes keys).  Lets replace them with registry that uses the format above, so this:

```
- type: REGISTRY_KEY
  attributes:
    keys:
      - 'HKEY_USERS\%%users.sid%%\Software\Microsoft\Windows\CurrentVersion\Run\*'
      - 'HKEY_USERS\%%users.sid%%\Software\Wow6432Node\Microsoft\Windows\CurrentVersion\Run\*'
```
becomes
```
- type: REGISTRY
  attributes:
    key_value_pairs:
      - {key: 'HKEY_USERS\%%users.sid%%\Software\Microsoft\Windows\CurrentVersion\Run\*'}
      - {key: 'HKEY_USERS\%%users.sid%%\Software\Wow6432Node\Microsoft\Windows\CurrentVersion\Run\*'}
```

destijl: the tricky thing there is how do you target the default value explicitly?  We could make a magic value like "@" or "Default" but that creates a problem if that's the actual string you want to target.  I think we need another parameter in this dict:

```
- type: REGISTRY
  attributes:
    key_value_pairs:
      - {key: 'HKEY_USERS\%%users.sid%%\Software\Microsoft\Windows\CurrentVersion\Run', default_value: true}
      - {key: 'HKEY_USERS\%%users.sid%%\Software\Wow6432Node\Microsoft\Windows\CurrentVersion\Run\*'}
```

destijl: If we don't do an explicit parameter I can't see how you could specify that you want the default value from a globbed set of keys like 'Windows\CurrentVersion\**'

pidydx: - type: REGISTRY
  attributes:
    path:
      - {key: 'HKEY_USERS\%%users.sid%%\Software\Microsoft\Windows\CurrentVersion\Run', value_name: svchost}
      - {key: 'HKEY_USERS\%%users.sid%%\Software\Microsoft\Windows\CurrentVersion\Run'}  #get default value
      - {key: 'HKEY_USERS\%%users.sid%%\Software\Wow6432Node\Microsoft\Windows\CurrentVersion\Run\*'}

destijl: Yeah, you're right, always returning the default value for any matching key is probably fine.

joachimmetz:
> Lets replace them with registry that uses the format above, so this:

The meaning of key and value are different. If I define a value I'm very likely interested in the content. When I define key I'm very likely interested in value or subkey names.

test_data/definitions.yaml
pidydx:
```
- type: REGISTRY
  attributes:
    path:
      - {value: 'HKEY_USERS\%%users.sid%%\Software\Microsoft\Windows\CurrentVersion\Run\svchost'}
      - {value: 'HKEY_USERS\%%users.sid%%\Software\Microsoft\Windows\CurrentVersion\Run'}  #get default value
      - {key: 'HKEY_USERS\%%users.sid%%\Software\Microsoft\Windows\CurrentVersion\Run'} #get all subkeys/values
      - {key: 'HKEY_USERS\%%users.sid%%\Software\Microsoft\Windows\CurrentVersion\Run\*'} #get all subkeys/values.  I would expect this to work the same as the previous entry barring dealing with the problems in using *
```

destijl: I think the value syntax there is too confusing, mixing keys and values didn't work out well the first time.

MSFT uses empty string for the default value:
https://msdn.microsoft.com/en-us/library/microsoft.win32.registrykey.getvaluenames(v=vs.110).aspx

I think we should just do that.

{key: 'CurrentVersion\Run\*' value: ''}  #get default value

destijl:
@joachimmetz agreed, but 
```
REGISTRY_KEY
  keys: "blah"
```
and
```
REGISTRY_VALUE:
   paths:
     - {key:"blah"}
```
are essentially two ways to write the same thing, and indicating you aren't interested in the value.  One format that is clear about keys and values is better I think.

destijl:
changed my mind, I think joachim's right, lets keep them separate.

artifacts/artifact.py
```diff
@@ -19,36 +18,42 @@ def __init__(self, name, description=None):
...
     self.name = name
     self.labels = []
     self.provides = []
+    self.sources = []
     self.supported_os = []
     self.urls = []
 
   # Property for name compatibility with current variant of GRR artifacts.
   @property
+  def collectors(self):
```
destijl: this should be sources?

artifacts/reader.py
```diff
@@ -149,4 +186,40 @@ def Read(self, file_object):
...
+  def ReadFile(self, filename):
+    """Reads artifact definitions from a YAML file.
+
+    Args:
+      filename: the name of the file to read from.
+
+    Yields:
+      Artifact definitions (instances of ArtifactDefinition).
+    """
+    file_object = open(filename, 'rb')
```
destijl: use the context manager to handle file open/close.

destijl:
I've got this format working in GRR. Can you sync this PR with head?

FYI here's the way I converted the artifacts for testing.  I'll post this in the docs once this change is in, so others can convert artifacts they are working on, or private artifacts.

```
sed -i s/collectors:/sources:/g *.yaml
sed -i s/collector_type:/type:/g *.yaml
sed -i s/path_list:/paths:/g *.yaml
sed -i s/args:$/attributes:/g *.yaml
sed -i s/args:\ {/attributes:\ {/g *.yaml
perl -0777 -i -pe "s/REGISTRY_KEY\n\ \ attributes:\n\ \ \ \ paths:/REGISTRY_KEY\n\ \ attributes:\n\ \ \ \ keys:/igs" windows_bootstrap.yaml windows.yaml
perl -0777 -i -pe "s/REGISTRY_KEY\n\ \ attributes:\ {paths:/REGISTRY_KEY\n\ \ attributes:\ {keys:/igs" webbrowser.yaml

sed -i "s/paths: \[\('HKEY[^']*'\)/paths: \[{key: \1}/g" windows.yaml windows_bootstrap.yaml
sed -i "s/paths: \[\('%%curr[^']*'\)/paths: \[{key: \1}/g" windows.yaml windows_bootstrap.yaml

perl -0777 -i -pe "s/REGISTRY_VALUE\n\ \ attributes:\ {paths:\ \[{key: '([^']*)\\\([^\']*)'}\]/REGISTRY_VALUE\n\ \ attributes:\ {key_value_pairs:\ [{key: '\1', value: '\2'}]/igs" windows_bootstrap.yaml windows.yaml
```

joachimmetz:
Thanks, I'll go over the comments and sync with head.

artifacts/artifact.py
joachimmetz: This initially was for backwards compatibility but I can remove this.

artifacts/reader.py
joachimmetz: ack.

joachimmetz:
CL updated.

destijl:
I just merged sean's PR, can you re-merge and fix those artifacts (you can use the commands above).  Then I think we're OK for this to go in.

joachimmetz:
Ok I'll re-merge and then merge the PR.

PR #12, opened 2015-01-30
Updated Prefetch to gather prefetch files generated by binaries executed...

pidydx:
... from ADS

### DIFF
--- a/definitions/windows.yaml
+++ b/definitions/windows.yaml
@@ -281,6 +281,15 @@ labels: [Users]
 supported_os: [Windows]
 urls: ['https://github.com/libyal/winreg-kb/wiki/Registry-files']
 ---
+name: WMIComputerSystemProduct
+doc: Computer System Product including Identifiying number queried from WMI.
+collectors:
+- collector_type: WMI
+  args: {query: SELECT * FROM Win32_ComputerSystemProduct}
+labels: [System]
+supported_os: [Windows]
+urls: ['http://msdn.microsoft.com/en-us/library/aa394105(v=vs.85).aspx']
+---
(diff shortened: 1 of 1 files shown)


joachimmetz:
This change makes the definitions too large.

What additional files do you think need to be added to the Prefetch files definition?

If you mean the *.db files those are likely [SuperFetch databases](https://github.com/libyal/libagdb/blob/master/documentation/Windows%20SuperFetch%20(DB)%20format.asciidoc) files not Prefetch files.

pidydx:
If I attach evil.exe as an ADS c:\windows\system32\notepad.exe:evil.exe then the resulting file will be ..\Prefetch\notepad.exe:evil.pf.  The same is true for directories.  It will create a file/directory for whatever the ADS was attached to and then create the .pf file as an ADS on that file/directory.

joachimmetz:
Yes but that would mean the glob *.pf applies as well. Not sure why you want to turn it into a * instead?

pidydx:
What would be the best way to define that in this artifact?

joachimmetz:
Let me sync with Greg and see if we can come up with a good solution. I'm leaning towards adding path with data stream/fork name or equiv.

joachimmetz:
I've created https://github.com/ForensicArtifacts/artifacts/issues/13 to track the status of this.

joachimmetz:
I'll will change the path definition to provide for a data stream name glob. The GRR implementation will initially not support this, but we'll work to get that support in.

Regarding this pull request, you should be able to create a custom artifact in GRR e.g. "PrefetchDirectory" as a work around.

PR #15, opened 2015-02-19
Fixed config_files.yaml separator

pidydx:


joachimmetz:
Thx that one slipped to the crack, I'll add an issue to set up travis to catch these.

PR #19, opened 2015-03-05
Added artifacts for kernel modules/extensions

pidydx:


definitions/darwin.yaml
```diff
@@ -80,4 +80,25 @@ collectors:
 - collector_type: FILE
   args: {path_list: ['%%users.homedir%%/Library/Preferences/*']}
 labels: [Users]
+---
```
joachimmetz: separate is in the wrong place

joachimmetz:
Please update with origin master and fix the separator.

pidydx:
This should be good now

PR #24, opened 2015-03-05
Added more Linux scheduled tasks

pidydx:


pidydx:
I modified this since some of the files under SystemCronFiles are crontabs that can be parsed with the parser I created for GRR and others are just scripts related to Anacron.  In non-debian systems there is an /etc/anacrontab that is formatted differently that drives the /etc/cron.(timeperiod)/* scripts.

destijl:
There's an /etc/anacron on ubuntu that does the same.  This LGTM.

pidydx:
Interesting.  I wonder if my Ubuntu VM is older then.  It was using /etc/crontab to start anacron /etc/cron.daily and so on.  I will try to dig into this a bit more later and maybe add some documentation.

PR #26, opened 2015-03-12
Add some proc/sys linux artifacts

destijl:


destijl:
@joachimmetz

destijl:
The best distinction I can make as to why these particular bits of proc, is that they are all security relevant.  To be more specific we'd need to break them all up individually, maybe that makes more sense....

sebastianwelsh:
See pull/27 for an updated set of proc entries with a more specific breakdown.

joachimmetz:
I'll have a look later today.

destijl:
closing this in favour of sebastian's version.

PR #27, opened 2015-03-13
Add some filesystem, kernel and network sysctl artifacts for Linux.

sebastianwelsh:
Adds security-relevant sysctl artifacts for linux.


definitions/linux.yaml
```diff
@@ -137,6 +137,116 @@ labels: [Authentication, Configuration Files]
...
+supported_os: [Linux]
+---
+name: ProcSysFsSecurity
+doc: Filesystem hardening settings.
+collectors:
+- collector_type: FILE
+  args:
+    path_list:
+      - '/proc/sys/fs/protected_hardlinks'
+      - '/proc/sys/fs/protected_symlinks'
```
destijl: hardlinks and symlinks should be grouped together since they relate to the same thing.  the suid_dumpable should be split out.

definitions/linux.yaml
```diff
@@ -137,6 +137,116 @@ labels: [Authentication, Configuration Files]
...
+labels: [System]
+supported_os: [Linux]
+---
+name: ProcSysKernelMemoryPrediction
+doc: Memory address obfuscation settings.
+collectors:
+- collector_type: FILE
+  args:
+    path_list:
+      - '/proc/sys/kernel/dmesg_restrict'
```
destijl: from a quick read these are only very loosely related.  I think all of these will be much more generally useful if we break them down individually with better names.  I can guess what dmesg_restrict is about, but if it was in it's own artifact called "DmesgReadPrivilegesRestriction" I wouldn't have to.  It would also be good to have canonical URL references for each of these so I don't have to go searching.

definitions/linux.yaml
```diff
@@ -137,6 +137,116 @@ labels: [Authentication, Configuration Files]
...
+labels: [System]
+supported_os: [Linux]
+---
+name: ProcSysKernelModules
+doc: Module loading options and taint state.
+collectors:
+- collector_type: FILE
+  args:
+    path_list:
+      - '/proc/sys/kernel/kexec_load_disabled'
```
destijl: Same goes for all the rest of these.  They should only be one artifact if they are very tightly related like: protected_hardlinks and symlinks; and
bootloader_type and version.

sebastianwelsh:
PTAL. Each artifact covers a single issue, more or less.

Also, I've stripped out the naming preamble for each, but if you want a prefix (e.g. Sysctl) I can do that. Given the number of artifacts, do you prefer a separate artifact file, or is it OK to include this in linux.yaml?

definitions/linux.yaml
```diff
@@ -137,6 +137,179 @@ labels: [Authentication, Configuration Files]
...
+collectors:
+- collector_type: FILE
+  args:
+    path_list:
+      - '/proc/sys/kernel/randomize_va_space'
+labels: [System]
+supported_os: [Linux]
+---
+name: IgnoreICMPBroadcasts
+doc: Whether the systrem ignores ICMP pings.
```
joachimmetz: system

definitions/linux.yaml
```diff
@@ -137,6 +137,179 @@ labels: [Authentication, Configuration Files]
...
+supported_os: [Linux]
+---
+name: KernelModuleProvenance
+doc: Taint state of loaded modules (binary blobs, unsigned modules etc).
+collectors:
+- collector_type: FILE
+  args:
+    path_list:
+      - '/proc/sys/kernel/tainted'
+labels: [System]
```
joachimmetz: + urls: ['https://www.kernel.org/doc/Documentation/sysctl/kernel.txt']

definitions/linux.yaml
```diff
@@ -137,6 +137,179 @@ labels: [Authentication, Configuration Files]
...
+supported_os: [Linux]
+---
+name: KernelModuleRestrictions
+doc: Module loading options and taint state.
```
destijl: no longer taint?

definitions/linux.yaml
```diff
@@ -137,6 +137,179 @@ labels: [Authentication, Configuration Files]
...
+supported_os: [Linux]
+---
+name: RestrictedKernelPointerReadPrivileges
+doc: Memory address obfuscation settings.
+collectors:
+- collector_type: FILE
+  args:
+    path_list:
+      - '/proc/sys/kernel/kptr_restrict'
+      - '/proc/sys/kernel/randomize_va_space'
```
destijl: you already have this one in ASLREnabled

destijl:
This looks much better. I think a new file makes sense, and /proc is a logical grouping, maybe linux_proc.yaml.

definitions/linux_proc.yaml
```diff
@@ -0,0 +1,186 @@
+# Linux specific /proc artifacts.
+name: ASLREnabled 
+doc: Kernel ASLR state.
+urls: ['https://www.kernel.org/doc/Documentation/sysctl/kernel.txt']
```
destijl: minor nit: put the urls last in the definition.

definitions/linux_proc.yaml
```diff
@@ -0,0 +1,186 @@
...
+urls: ['https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt']
+collectors:
+- collector_type: FILE
+  args:
+    path_list:
+      - '/proc/sys/net/ipv4/tcp_syncookies'
+labels: [Network, System]
+supported_os: [Linux]
+---
+name: ProcSys
```
destijl: LinuxProcSysHardeningSettings is probably a better name to describe this particular subset of /proc/sys

destijl:
Looks great, yeah I think we should prefix all the Linux ones with Linux.  Just change this file for now, we'll need to coordinate a change with grr for the others with existing parsers.

definitions/linux_proc.yaml
sebastianwelsh: The bootloader entry is more informational, but otherwise, sure.

sebastianwelsh:
PTAL

definitions/linux_proc.yaml
```diff
@@ -0,0 +1,187 @@
...
+- collector_type: FILE
+  args:
+    path_list:
+      - '/proc/sys/kernel/bootloader_type'
+      - '/proc/sys/kernel/bootloader_version'
+labels: [System]
+supported_os: [Linux]
+urls: ['https://www.kernel.org/doc/Documentation/sysctl/kernel.txt']
+---
+name: LinuxKernelModuleProvenance
```
destijl: LinuxKernelModuleTaintStatus

destijl:
One tiny change, LGTM.  Make that and I'll merge.

sebastianwelsh:
et voilà!

destijl:
Thanks for all the work here.

PR #30, opened 2015-03-19
Fix minor lint issues, remove unnecessary kwargs

destijl:
The kwargs used by subclasses was weird.  If they were ever set we just raised, so having it seems pretty pointless.

@joachimmetz


destijl:
OK, I'll rethink this.

destijl:
As is, you essentially modify the API to allow for failure then check for the failure.  But this is unnecessary. since the classes already define what arguments they can accept.  I added a test to demonstrate. PTAL.

joachimmetz:
If not needed I'm fine with getting rid of it.

destijl:
OK, merging now.

PR #32, opened 2015-03-27
Added some shell artifacts and moved shared unix shell related artifacts...

pidydx:
... to a combined location

### DIFF
--- a/definitions/shells.yaml
+++ b/definitions/shells.yaml
@@ -1,0 +1,106 @@
+name: RootUserShellConfigs
+doc: Linux root shell configuration files.
+collectors:
+- collector_type: FILE
+  args:
+    path_list:
+      - '/root/.bashrc'
+      - '/root/.bash_profile'
+      - '/root/.bash_logout'
+      - '/root/.cshrc'
+      - '/root/.ksh'
+      - '/root/.logout'
+      - '/root/.profile'
+      - '/root/.tcsh'
+      - '/root/.zlogin'
+      - '/root/.zlogout'
+      - '/root/.zprofile'
+labels: [Configuration Files]
+supported_os: [Linux]
+---
+name: UsersShellConfigs
+doc: Linux root shell configuration files.
+collectors:
+- collector_type: FILE
+  args:
+    path_list:
+      - '%%users.homedir%%/.bashrc'
+      - '%%users.homedir%%/.bash_profile'
+      - '%%users.homedir%%/.bash_logout'
+      - '%%users.homedir%%/.cshrc'
+      - '%%users.homedir%%/.ksh'
+      - '%%users.homedir%%/.logout'
+      - '%%users.homedir%%/.profile'
+      - '%%users.homedir%%/.tcsh'
+      - '%%users.homedir%%/.zlogin'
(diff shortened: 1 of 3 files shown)


definitions/shells.yaml
```diff
@@ -0,0 +1,106 @@
...
+      - '/root/.profile'
+      - '/root/.tcsh'
+      - '/root/.zlogin'
+      - '/root/.zlogout'
+      - '/root/.zprofile'
+labels: [Configuration Files]
+supported_os: [Linux]
+---
+name: UsersShellConfigs
+doc: Linux root shell configuration files.
```
destijl: user

definitions/shells.yaml
```diff
@@ -0,0 +1,106 @@
...
+    path_list:
+      - '/root/.bash_history'
+      - '/root/.sh_history'
+      - '/root/.zhistory'
+      - '/root/.zsh_history'
+labels: [Configuration Files]
+supported_os: [Linux]
+---
+name: UsersShellHistory
+doc: Linux root shell configuration files.
```
destijl: user

definitions/shells.yaml
```diff
@@ -0,0 +1,106 @@
...
+supported_os: [Linux, Darwin]
+---
+name: UnixPasswd
+doc: Linux or OSX passwd file.
+collectors:
+- collector_type: FILE
+  args: {path_list: ['/etc/passwd']}
+labels: [Authentication]
+supported_os: [Linux, Darwin]
+---
```
destijl: this extra separator is probably causing the test failure.

definitions/shells.yaml
```diff
@@ -0,0 +1,106 @@
...
+      - '/etc/zsh/zlogout'
+      - '/etc/zsh/zprofile'
+      - '/etc/zsh/zshenv'
+      - '/etc/zsh/zshrc'
+      - '/etc/zshenv'
+      - '/etc/zshrc'      
+labels: [Configuration Files]
+supported_os: [Linux, Darwin]
+---
+name: SudoersConfiguration
```
destijl: these need to be removed from linux.yaml.  I feel like they should be in a separate file that covers the commonalities between os x and linux.  Maybe nix.yaml?

pidydx:
I think I got all the cleanup things you mentioned.  I went with unix_common.yaml for the shared artifacts.

definitions/unix_common.yaml
```diff
@@ -0,0 +1,111 @@
+# Common Unix based artifacts
```
destijl: I think this file should only be those things that are shared between linux and OS X.  Move the linux-only ones back into the linux file.

definitions/unix_common.yaml
```diff
@@ -0,0 +1,111 @@
...
+      - '/root/.profile'
+      - '/root/.tcsh'
+      - '/root/.zlogin'
+      - '/root/.zlogout'
+      - '/root/.zprofile'
+labels: [Configuration Files]
+supported_os: [Linux]
+---
+name: UsersShellConfigs
+doc: Common unix user shell configuration files.
```
destijl: And I think you can drop "Common"

PR #33, opened 2015-03-27
Added more Windows Artifacts

pidydx:
[Review on Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/33)

### DIFF
--- a/definitions/windows.yaml
+++ b/definitions/windows.yaml
@@ -199,7 +199,7 @@ doc: |
   Note users.sid will currently only expand to SIDs with profiles
   on the system, not all SIDs.
 collectors:
-- collector_type: REGISTRY_KEY
+- collector_type: REGISTRY_KEY #REGISTRY_KEY
   args:
     path_list:
       - 'HKEY_USERS\%%users.sid%%\Software\Microsoft\Windows\CurrentVersion\Run\*'
@@ -208,6 +208,7 @@ collectors:
       - 'HKEY_USERS\%%users.sid%%\Software\Wow6432Node\Microsoft\Windows\CurrentVersion\RunOnce\*'
       - 'HKEY_USERS\%%users.sid%%\Software\Microsoft\Windows\CurrentVersion\Policies\Explorer\Run\*'
       - 'HKEY_USERS\%%users.sid%%\Software\Wow6432Node\Microsoft\Windows\CurrentVersion\Policies\Explorer\Run\*'
+      - 'HKEY_USERS\%%users.sid%%\Software\Microsoft\Windows\CurrentVersion\RunOnce\Setup\*'
       - 'HKEY_LOCAL_MACHINE\Software\Microsoft\Windows\CurrentVersion\Run\*'
       - 'HKEY_LOCAL_MACHINE\Software\Wow6432Node\Microsoft\Windows\CurrentVersion\Run\*'
       - 'HKEY_LOCAL_MACHINE\Software\Microsoft\Windows\CurrentVersion\RunOnce\*'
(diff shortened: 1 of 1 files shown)


destijl:
Awesome, thanks! It would be really nice to have some documentation links though.

pidydx:
I have been trying to figure out the best way to handle that.  I might just dig up some MSDN links for things until I can come up with something better.

joachimmetz:
An alternative is adding documentation to forensicswiki.org

joachimmetz:
@pidydx can you make sure to mark Registry values different from Registry keys, I know the format currently does not support it but it makes life easier when it does

pidydx:
Re: Documentation:  I do have a forensicswiki.org account now, but I want to make sure I can find "official" documentation on the details of how some of the keys work so I don't wind up documenting screwed up descriptions of the persistence mechanisms or expected values.  My longer term plan is to sort all these things by loose categories of why someone would care about each one.  (Persistence, Escalation, History/Forensics, etc)

Re: Key/Value.  Do you mean to mark them by which ones are returning a single specific Value vs the ones that are going to return a set of Keys and values?

joachimmetz:
> I want to make sure I can find "official" documentation on the details of how some of the keys work so I don't wind up documenting screwed up descriptions of the persistence mechanisms or expected values. 

I've notices that sometimes this means piecing together various MSDN sources. Also a wiki article could give more room for actually describing Persistence, Escalation, History/Forensics, etc.

> Key/Value. Do you mean to mark them by which ones are returning a single specific Value vs the ones that are going to return a set of Keys and values?

In https://github.com/ForensicArtifacts/artifacts/pull/11 I'm separating Registry keys from values (since the current variant does not support this). If you can indicate in the doc string if its a key or value that could be useful when merging these with the pull request.

pidydx:
>I've notices that sometimes this means piecing together various MSDN sources. Also a wiki article could give more room for actually describing Persistence, Escalation, History/Forensics, etc.

The biggest headache I have run into is which HKU keys can override HKLM ones so I have a lot of artifacts that just attempt to grab both just to be sure.  My plan is to dig through MSDN, Windows Internals, and threat research blog stuff to try and put something reasonable together after I get done brain dumping all of my notes into artifacts.  Unfortunately most of my notes are "You care about this key" and less "and this is why".

>In #11 I'm separating Registry keys from values (since the current variant does not support this). If you can indicate in the doc string if its a key or value that could be useful when merging these with the pull request.

Ah, I see how you mean.  Should I just document or do you want me to change to the new format?

joachimmetz:
> Ah, I see how you mean. Should I just document or do you want me to change to the new format?

Since it is not yet part of HEAD I opt to just document, otherwise you'll have to wait for the change to be submitted before we can integrate any of your changes in the mean time.

joachimmetz:
> My plan is to dig through MSDN, Windows Internals, and threat research blog stuff to try and put something reasonable together after I get done brain dumping all of my notes into artifacts.

For reference URL I personally would go for forensicswiki (otherwise a github wiki if you want more control over it?), the advantage being that other can help you out writing. Side note: I'm personally not a fan of using blogs for reference material.

pidydx:
I just need to find the time to sit down and write up the forensic wiki stuff.  I don't generally trust blogs to stay online anyways, but I have a few links that can at least be temporary documentation of why a particular artifact is interesting.

joachimmetz:
That would be great, thanks in advance.

pidydx:
I added a few more and put comments in place for REGISTRY_VALUE vs REGISTRY_KEY.  I really like the idea of removing ambiguity like this as it has been a pain to get definitive answers on what some of the optional entries really are.  I can correct any I mislabeled when I can test more of them against the fleet.

joachimmetz:
>  I really like the idea of removing ambiguity like this as it has been a pain to get definitive answers on what some of the optional entries really are.

It also necessary since technically a value can:

* have the same name as a subkey e.g. BagMRU
* contain the Registry path separator `\` as a valid character

destijl:
we're pretty close to getting the format change PR in, I think you should wait on this one and just move to the new format since you have a bunch of reg key values that would benefit.

pidydx:
I was afraid of that.  Converting this unmerged mess is going to be a pain. :) I will try to get this cleaned up soon.

joachimmetz:
Not sure why you closed this. @pidydx should otherwise Greg or my try to merge these in?

pidydx:
Yes, this one was using the old format so rather than dealing with a crazy merge mess I just pulled the latest and rebuilt using the new format.

joachimmetz:
Great, was missing that context, I'll have a look at the new PR.

PR #39, opened 2015-04-09
Split unix passwd, shadow, group artifacts.

destijl:
@joachimmetz @pidydx 

GRR has a parser that works on /etc/passwd, so I need these split out.


pidydx:
LGTM

joachimmetz:
LTGM, one nit, why are they not using the short hand style as described here
https://github.com/ForensicArtifacts/artifacts/blob/master/docs/style_guide.adoc#sources

destijl:
You're right, I'll fix that now.

PR #40, opened 2015-04-09
Fix SystemDrive.

destijl:
This is not exactly ideal, but is less bad than the current state, which is broken.


destijl:
@pidydx

destijl:
@joachimmetz I'm going to put this in TBR, it's blocking me getting GRR working.  Discuss tomorrow.

joachimmetz:
LGTM, found this [MSDN](https://msdn.microsoft.com/en-us/library/cc231436.aspx) article that provides more context. I've added it in the dpkg pull request as a URL.

PR #45, opened 2015-04-21
Add xinetd configuration artifacts.

sebastianwelsh:
Let's collect xinetd service definitions as well.

### DIFF
--- a/definitions/linux.yaml
+++ b/definitions/linux.yaml
@@ -219,3 +219,14 @@ collectors:
 labels: [Configuration Files, System]
 supported_os: [Linux]
 urls: ['https://wiki.debian.org/LSBInitScripts']
+---
+name: LinuxXinetd
+doc: Linux xinetd configurations.
+sources:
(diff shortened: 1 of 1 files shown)


definitions/linux.yaml
```diff
@@ -219,3 +219,14 @@ collectors:
...
+name: LinuxXinetd
+doc: Linux xinetd configurations.
+sources:
+- type: FILE
+  attributes:
+    paths:
+      - '/etc/xinetd.conf'
+      - '/etc/xinetd.d/**'
+labels: [Configuration Files, System]
+supported_os: [Linux]
```
joachimmetz: + urls: ['http://en.wikipedia.org/wiki/Xinetd']

joachimmetz:
LGTM, please add documentation a URL, don't expect every consumer of the artifacts to know Linux that well

sebastianwelsh:
OK, URL to wikipedia added.

PR #49, opened 2015-04-30
Update AnacronFiles & LinuxCronTabs and add LinuxScheduleFiles

timevortex:
Include files inside the directories of cron.* Also /var/spool/cron contains further directories.

LinuxScheduleFiles combines all three scheduling artificats into one.

### DIFF
--- a/definitions/linux.yaml
+++ b/definitions/linux.yaml
@@ -7,10 +7,10 @@ sources:
   attributes:
     paths:
       - '/etc/anacrontab'
-      - '/etc/cron.daily'
-      - '/etc/cron.hourly'
-      - '/etc/cron.monthly'
-      - '/etc/cron.weekly'
+      - '/etc/cron.daily/*'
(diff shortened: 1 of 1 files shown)


definitions/linux.yaml
```diff
@@ -38,6 +38,16 @@ sources:
 labels: [Configuration Files]
 supported_os: [Linux]
 ---
+name: LinuxScheduleFiles
+doc: All Linux job scheduling files.
+sources:
+- type: ARTIFACT
+  attributes:
+    names: [AnacronFiles, LinuxCronTabs, LinuxAtJobs]
+  returned_types: [StatEntry]
```
destijl: Lets drop returned_types, unless there's a strong reason to keep it.  This is a GRR-ism I'd like to keep out of this repo as much as possible.

Otherwise, looks great, thanks!

timevortex:
Have removed returned_types =)

PR #50, opened 2015-04-30
Fix flaky test

destijl:


destijl:
@joachimmetz

artifacts/registry.py
```diff
@@ -49,8 +49,7 @@ def GetDefinitions(self):
     Yields:
```
joachimmetz: change the docstring accordingly

joachimmetz:
LGTM, one small docstring nit

artifacts/registry.py
destijl: done.

PR #52, opened 2015-05-07
Added more cron defintions

timevortex:
Added stating of parent folders already defined in existing artifiacts and also added collecting information about /etc/cron.allow, /etc/cron.deny, /etc/at.allow and /etc/at.deny.

[Review on Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/52)

### DIFF
--- a/definitions/linux.yaml
+++ b/definitions/linux.yaml
@@ -47,6 +47,22 @@ sources:
 labels: [Configuration Files]
 supported_os: [Linux]
 ---
+name: CronAtAllowDenyFiles
+doc: Files containing users authorised to run cron or at jobs.
+sources:
+- type: FILE
+  attributes:
+    paths:
+      - '/etc/cron.allow'
+      - '/etc/cron.deny'
+      - '/etc/at.allow'
+      - '/etc/at.deny'
+labels: [Configuration Files]
+supported_os: [Linux]
+urls:
+  - http://manpages.ubuntu.com/manpages/saucy/man5/at.allow.5.html
+  - http://manpages.ubuntu.com/manpages/precise/en/man1/crontab.1.html
+---
 name: HostAccessPolicyConfiguration
 doc: Linux files related to host access policy configuration.
 sources:


timevortex:
Tests failing because LIST_FILES not supported by validator but it's acceptable in grr: https://github.com/google/grr/blob/4e36bfa891a5749a76ea3cfc2b089099b470e38a/lib/artifact_lib.py#L79

>> Runs the validator over all the YAML artifact definitions files. ... WARNING:root:Invalid artifact definition: LinuxCronParentFolders. Unsupported type indicator: LIST_FILES.

timevortex:
I have made the necessary changes to add support for the LIST_FILES artifact so that the tests pass. I'm not sure if this is the best way to proceed forward (happy to take suggestions!) but at least the tests aren't failing now =)

TO DO: Add LIST_FILES as a source type in documentation: https://github.com/ForensicArtifacts/artifacts/blob/master/docs/Artifacts%20definition%20format%20and%20style%20guide.asciidoc#source-types. I'm not sure what the best description for LIST_FILES is. Happy to do this mod and include it in this pull request if someone suggests to me what I should put in the description.

joachimmetz:
LIST_FILES is a GRR flow and should not be an artifact source type.
GRR maps the artifacts to flows actions.

Artifacts are descriptive by design (as data) not functional (as code).

You likely want the FILE source type: https://github.com/ForensicArtifacts/artifacts/blob/master/docs/Artifacts%20definition%20format%20and%20style%20guide.asciidoc#file-source

timevortex:
FILE get contents of files and hence does not include stat of directories but only files.

Hence why I created LinuxCronParentFolders - I want the permissions on those folders but FILE will not give them to me only LIST_FILES will in grr.

joachimmetz:
FILE has no logic to it, it does nothing by itself; it is only descriptive. Logic is not portable and the idea of artifacts is to reuse the same data in different contexts e.g. GRR, plaso, TimeSketch, who knows.

The functionality you want should be implemented in GRR, the artifacts themselves only tell GRR what the source location of the files are. They don't tell GRR what information to collect. Including @destijl he might be able to point you to the right direction in GRR to accomplish what you want to achieve.

destijl:
Hey @timevortex joachim is right.  The distinction we're making between GRR and this repo is that stuff in this repo describes where things are on the machine, but not how they should be collected (since this is tool specific and varies wildly).

However, we're using artifacts in GRR as a simple way to programmatically call grr-specific flows, called flow_templates for the time being, so this should go in here:
https://github.com/google/grr/blob/master/artifacts/flow_templates/linux.yaml

destijl:
Reviewed files:
- artifacts/artifact.py @ r2
- artifacts/definitions.py @ r2
- artifacts/source_type.py @ r3
- definitions/linux.yaml @ r1

---

<sup>**[definitions/linux.yaml, line 67 \[r3\]](https://reviewable.io:443/reviews/forensicartifacts/artifacts/52#-JopNy0yHDM0khA3kux4)** ([raw file](https://github.com/forensicartifacts/artifacts/blob/ddaf0161c09854393c30a3ef80b84630ceefb7a0/definitions/linux.yaml#L67)):</sup>
this artifact looks good, drop the rest out and put just this bit in.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/forensicartifacts/artifacts/52)

timevortex:
Hi @destijl and @joachimmetz I have reverted all the other changes and just left CronAtAllowDenyFiles as an addition in linux.yaml

destijl:
Thanks!

PR #64, opened 2015-05-28
Collect /proc/mounts and /etc/fstab artifacts together.

sebastianwelsh:
Move LinuxProcMounts out of the LinuxProcSysHardeningSettings

[Review on Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/64)

### DIFF
--- a/definitions/linux.yaml
+++ b/definitions/linux.yaml
@@ -163,6 +163,16 @@ labels: [System, Configuration Files]
 supported_os: [Linux]
 urls: ['http://en.wikipedia.org/wiki/Fstab']
 ---
+name: LinuxMountInfo
+doc: Linux mount options.
+sources:
+- type: ARTIFACT
+  attributes:
+    names: [LinuxFstab, LinuxProcMounts]
+labels: [System, Configuration Files]
+supported_os: [Linux]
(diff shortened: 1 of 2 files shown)


definitions/linux.yaml
```diff
@@ -163,6 +163,16 @@ labels: [System, Configuration Files]
 supported_os: [Linux]
 urls: ['http://en.wikipedia.org/wiki/Fstab']
 ---
+name: LinuxMountInfo
+doc: Linux mount options.
+sources:
+- type: ARTIFACT
+  attributes:
+    names: [LinuxFstab, LinuxProcMounts]
```
joachimmetz: Use long form
names:
`-` LinuxFstab
...

definitions/linux.yaml
```diff
@@ -163,6 +163,16 @@ labels: [System, Configuration Files]
...
+labels: [System, Configuration Files]
+supported_os: [Linux]
+urls: ['http://en.wikipedia.org/wiki/Fstab']
```
joachimmetz: - URL
since that is already defined in the LinuxFstab artifact itself

sebastianwelsh:
PTAL

PR #71, opened 2015-06-17
Multiple configs for syslog-ng

timevortex:
Not covered in the man page but seems to be utilised on at least Debian/Ubuntu by using a @include "/etc/syslog-ng/conf.d/*.conf" at the end of the primary syslog-ng.conf file.

Sorry for the unnecessary overhead :(

[Review on Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/71)


joachimmetz:
LGTM

PR #75, opened 2015-06-22
Remove users.local_settings

destijl:
AFAICT local settings is not actually a configurable environment variable, and
is always:

%USERPROFILE%\Local Settings

on the versions of windows where it existed. We never actually populated this
variable in GRR, so I'm going to remove it there as well. This Feeds Cache IE
history location appears to be only for very old versions of windows/IE and
isn't actually listed in the forensicswiki reference.

[Review on Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/75)


destijl:
@joachimmetz, relevant GRR bug is https://github.com/google/grr/issues/197

joachimmetz:
LGTM

PR #79, opened 2015-06-25
Add pam file and ps command artifacts.

sebastianwelsh:
[Review on Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/79)

### DIFF
--- a/definitions/linux.yaml
+++ b/definitions/linux.yaml
@@ -180,6 +180,19 @@ sources:
 labels: [Authentication, Configuration Files]
 supported_os: [Linux]
 ---
+name: LinuxPamConfigs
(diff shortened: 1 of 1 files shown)


definitions/linux.yaml
```diff
@@ -345,6 +358,16 @@ sources:
 labels: [Configuration Files, System]
 supported_os: [Linux]
 ---
+name: ListProcessesPs
```
destijl: Lets call it ListProcessesPsCommand so it's easily differentiated from in the GRR action one.

destijl:
LGTM, just a minor change

sebastianwelsh:
PTAL

PR #86, opened 2015-08-26
Add several more Windows artifacts.

Onager:
Mostly related to different persistence techniques.

[Review on Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/86)

### DIFF
--- a/definitions/windows.yaml
+++ b/definitions/windows.yaml
@@ -95,6 +95,20 @@ sources:
 supported_os: [Windows]
 urls: ['http://dfrws.org/2015/proceedings/presentations/DFRWS2015-pres3.pdf']
 ---
+name: EnvironmentUserLoginScripts
+doc: User login scripts configured via environment variables.
+source:
+- type: REGISTRY_VALUE
+  attributes:
+    key_value_pairs:
+        - {key: 'HKEY_USERS\%%users.sid%%\Environment', value: 'UserInitLogonServer'}
+        - {key: 'HKEY_USERS\%%users.sid%%\Environment', value: 'UserInitLogonScript'}
+        - {key: 'HKEY_USERS\%%users.sid%%\Environment', value: 'UserMprLogonScript'}
+supported_os: [Windows]
+urls:
+- 'http://www.hexacorn.com/blog/2014/11/14/beyond-good-ol-run-key-part-18/'
+- 'https://social.msdn.microsoft.com/Forums/windowsdesktop/en-US/cb6f1d6f-60a6-4369-803e-ec03d902e638/gina-how-to-run-domain-scripts-after-logon'
+---
 name: EventLogs
 doc: Windows Event logs.
 sources:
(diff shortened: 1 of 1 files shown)


definitions/windows.yaml
```diff
@@ -24,6 +24,31 @@ labels: [Logs]
...
+    - 'HKEY_LOCAL_MACHINE\\Software\\Microsoft\\Windows\\CurrentVersion\\App Paths\\*'
+    - 'HKEY_LOCAL_MACHINE\\Software\\Wow6432Node\\Microsoft\\Windows\\CurrentVersion\\App Paths\\*'
+supported_os: [Windows]
+urls:
+  - 'https://github.com/keydet89/RegRipper2.8/blob/master/plugins/apppaths.pl'
+  - 'http://www.hexacorn.com/blog/2013/01/19/beyond-good-ol-run-key-part-3/'
+  - 'https://msdn.microsoft.com/en-us/library/windows/desktop/ee872121(v=vs.85).aspx'
+---
+name: APPX_PROCESSEnvironmentVariable
+doc: APPX_PROCESS environment variable.
```
destijl: This is awesome, thanks. All of these would be much more usable if they had a one line summary in the doc that gives you the essential context without having to read a giant blogpost, or 3 giant blogposts.  e.g for this one:

```If this environment variable is set, .NET applications will attempt to load WinAppXRT.dll from PATH which is a potential persistence mechanism.```

I can help out by writing a few of them if you like.

File: definitions/windows.yaml [hunk 2]
```diff
@@ -57,6 +82,17 @@ sources:
 labels: [Logs]
 supported_os: [Windows]
 ---
+name: RunGrpConvFlag
```
destijl: put this in alphabetical order.

definitions/windows.yaml
```diff
@@ -645,6 +755,20 @@ sources:
   attributes: {query: SELECT * FROM Win32_Service}
 supported_os: [Windows]
 ---
+name: WinAppXRT
+doc: WinAppXRT DLL loaded when the APPX_PROCESS environment variable is set.
+sources:
+- type: FILE
+  attributes:
+    paths:
+    - '%%environ_systemroot%%\system32\WinAppXRT.dll'
```
destijl: This is OK for now, but it would be even better if you could just do:

```%%environ_path%%\\WinAppXRT.dll```

and have GRR collect environ_path (which it already does) and then figure out all the potential path combinations based on that (which it doesn't do). Filed https://github.com/google/grr/issues/250

File: definitions/windows.yaml (hunk 1 above)
Comment (Onager) [inline definitions/windows.yaml]: I can do this - I was trying to follow the style of the other doc strings, which seemed fairly terse.

**Comment by destijl** [comment]
Still happy to help out here if you're having trouble fitting this in, let me know.

File: definitions/windows.yaml (hunk 1 above)
Comment (Onager) [inline definitions/windows.yaml]: This is done now.

File: definitions/windows.yaml (hunk 2 above)
Comment (Onager) [inline definitions/windows.yaml]: Done.

File: definitions/windows.yaml [hunk 4]
```diff
@@ -237,6 +336,20 @@ labels: [Logs]
 supported_os: [Windows]
 urls: ['http://www.forensicswiki.org/wiki/Windows_XML_Event_Log_(EVTX)']
 ---
+name: EnvironmentUserLoginScripts
```
destijl: Put in alphabetic order.

destijl:
LGTM, just one minor change.

definitions/windows.yaml
Onager: Done.

Issue #88, opened 2015-08-28
exception.message is not portable to Python 3


joachimmetz:
https://github.com/ForensicArtifacts/artifacts/blob/6b789d6cddec1304f26e466eb19905468b151de7/tools/validator.py#L44

PR #90, opened 2015-09-20
Fix for Python 3 compatibility #88

joachimmetz:
@destijl for review

[Review on Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/90)

### DIFF
--- a/tools/validator.py
+++ b/tools/validator.py
@@ -42,7 +42,9 @@ class ArtifactDefinitionsValidator(object):
           result = False
 
     except errors.FormatError as exception:
-      logging.warning(exception.message)
+      logging.warning(
+          u'Unable to validate file: {0:s} with error: {1:s}'.format(
+              filename, exception))
       result = False
 
     return result


tools/validator.py
```diff
@@ -42,7 +42,9 @@ def CheckFile(self, filename):
           result = False
 
     except errors.FormatError as exception:
-      logging.warning(exception.message)
+      logging.warning(
+          u'Unable to validate file: {0:s} with error: {1:s}'.format(
```
destijl: you can remove the :s here and on line 40, it is the default.

destijl:
LGTM, just a minor nit.

tools/validator.py
joachimmetz: I rather not, I like code to be explicit about format indicators.

destijl: I think it's ugly. Since you almost always want a string conversion just being explicit when it *isn't* a string makes sense to me, but leave it if you like.

joachimmetz: Ack. Note tjat we use the explicit notation across the rest of the codebase so I'm sticking with keeping it to match that style and not to mix styles.

PR #92, opened 2015-09-23
Added Windows Google Drive client artifacts.

joachimmetz:
@destijl for review

Note that I'll need to create the forensics wiki page and add Mac OS X at a later date

[Review on Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/92)

### DIFF
--- a/definitions/cloud_services.yaml
+++ b/definitions/cloud_services.yaml
@@ -1,0 +1,27 @@
+# Cloud service artifacts.
+
+name: CloudStorageClients
+doc: Multiple cloud storage client artifacts.
+sources:
+- type: ARTIFACT
+  attributes:
+    names:
+      - 'GoogleDriveClient'
+labels: [Cloud Storage]
+supported_os: [Darwin,Linux,Windows]
+---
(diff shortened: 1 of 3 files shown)


definitions/cloud_services.yaml
```diff
@@ -0,0 +1,27 @@
+# Cloud service artifacts.
+
+name: CloudStorageClients
```
destijl: This is confusing when there is only one here, lets just create it when we have more than one.

destijl:
LGTM

definitions/cloud_services.yaml
joachimmetz: True, let me see if should add some more ;)

joachimmetz:
@destijl can you have a second look at this I've added dropbox as well to make sure the CloudStorageClients had more than one entry.

destijl:
LGTM but lets hold off on merging this until we can update GRR with the new label, or it will break if people import it into GRR.

destijl:
The new label is present in grr, so this can go in.

PR #98, opened 2015-10-13
OSX - Provide a flow collecting mounted DMG files info.

miloszlakomy:
Add an artifact that runs `hdiutil info` and collects its output.

This commit depends on a client template change that's on its way
to be synced into github.com/google/grr at the time of this writing.

[Review on Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/98)


destijl:
Thanks!

PR #101, opened 2015-10-15
Fixed syntax for GRR usage

pidydx:
This broke GRR Interrogate.

[Review on Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/101)


destijl:
Thanks, I'll look at what happened to the validation here....

PR #105, opened 2015-10-19
Add extra validation.

destijl:
This catches the sources syntax error we saw in
https://github.com/ForensicArtifacts/artifacts/pull/101.

[Review on Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/105)

### DIFF
--- a/artifacts/reader.py
+++ b/artifacts/reader.py
@@ -74,7 +74,8 @@ class YamlArtifactsReader(ArtifactsReader):
 
     different_keys = set(yaml_definition) - definitions.TOP_LEVEL_KEYS
     if different_keys:
-      raise errors.FormatError(u'Bad keys: {0!s}'.format(different_keys))
+      raise errors.FormatError(u'Undefined keys: {0!s}'.format(
+          different_keys))
 
     name = yaml_definition.get(u'name', None)
     if not name:
--- a/artifacts/reader.py
+++ b/artifacts/reader.py
@@ -72,6 +72,10 @@ class YamlArtifactsReader(ArtifactsReader):
     if not yaml_definition:
       raise errors.FormatError(u'Missing YAML definition.')
 
+    different_keys = set(yaml_definition) - definitions.TOP_LEVEL_KEYS
+    if different_keys:
+      raise errors.FormatError(u'Bad keys: {0!s}'.format(different_keys))
+
     name = yaml_definition.get(u'name', None)
     if not name:
       raise errors.FormatError(u'Invalid artifact definition missing name.')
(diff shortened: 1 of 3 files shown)


artifacts/reader.py
```diff
@@ -72,6 +72,10 @@ def _ReadArtifactDefinition(self, yaml_definition):
     if not yaml_definition:
       raise errors.FormatError(u'Missing YAML definition.')
 
+    different_keys = set(yaml_definition) - definitions.TOP_LEVEL_KEYS
+    if different_keys:
+      raise errors.FormatError(u'Bad keys: {0!s}'.format(different_keys))
```
joachimmetz: Bad keys => Undocumented keys: {0:s} ...

... check with project maintainers (or equiv?)

joachimmetz: So the format spec is currently not strict about having additional keys. I opt to have GRR be more tolerant in handling these but I'm fine with having a "strict" format check. However we should not point out these "keys" as "bad", since they are merely undocumented.

destijl: The intention here is to catch mistyped optional keys like "urps" instead of "urls". This doesn't signal a reluctance to add additional keys in the future, just that they should be strictly defined in the spec so we can do proper validation. I think it's fine to describe any key not in the spec as "bad", but I'll change the error text here if you like. This also isn't a format change so I don't think there's any need to discuss with others. If people want to have private artifacts with their own custom keys they should override TOP_LEVEL_KEYS to include them and get proper validation.

{0:s} doesn't work for sets, you need to do the str() call first, which is what {0!s} does. Again, my preference is not to have the explicit conversion for strings, the language isn't designed to require it, and string conversion is almost always what you want. If we weren't being explicit it would just work.

joachimmetz: > The intention here is to catch mistyped optional keys like "urps" instead of "urls".

I see, it was not clear to me from the description in https://github.com/google/grr/issues/277 that it was caused by a typo.

> {0:s} vs. {0!s}

I meant to use {0!s} in the example.

> Again, my preference is not to have the explicit conversion for strings

If we want to change it, that fine with me if we do it for the whole codebase at once. I would opt that we don't mix styles.

> I'll change the error text here if you like

Then I opt:
```
raise errors.FormatError(u'Unsupported keys: {0!s}'.format(different_keys))
```

And remove the {0:s} type notation in a follow up CL.

> This also isn't a format change so I don't think there's any need to discuss with others.

I meant this as part of the error message, when the error is raised to indicate changes might need to be added to the format.

joachimmetz:
LGTM

PR #106, opened 2015-10-23
update RunGrpConv keys to be a list

liamjm:
Reviewer: @destijl


destijl:
Thanks, looks like we need another validation code improvement.

PR #113, opened 2015-10-30
Add validation to Win Registry Key prefixes

liamjm:
Add some extra checks to the reading of windows Registry Keys
* ensure keys is a list for TYPE_INDICATOR_WINDOWS_REGISTRY_KEY
* for each key, ensure it starts with an acceptable prefix (don't us HKLM for example)

Follow-on tweaks to some of the definitions to pass the new verification.

Reviewer: @destijl, @joachimmetz

[Review on Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/113)

### DIFF
--- a/artifacts/source_type.py
+++ b/artifacts/source_type.py
@@ -165,6 +165,13 @@ class WindowsRegistryKeySourceType(SourceType):
 
   TYPE_INDICATOR = definitions.TYPE_INDICATOR_WINDOWS_REGISTRY_KEY
 
+  VALID_PREFIXES = [
+      r'HKEY_LOCAL_MACHINE',
+      r'HKEY_USERS',
+      r'HKEY_CLASSES_ROOT',
+      r'%%current_control_set%%',
+  ]
+
   def __init__(self, keys=None):
     """Initializes the source type object.
 
@@ -178,8 +185,21 @@ class WindowsRegistryKeySourceType(SourceType):
     if not keys:
       raise errors.FormatError(u'Missing keys value.')
 
+    if not isinstance(keys, list):
+      raise errors.FormatError(u'keys must be a list')
+
+    for key in keys:
+      self.ValidateKey(key)
+
     super(WindowsRegistryKeySourceType, self).__init__()
     self.keys = keys
+
+  @classmethod
+  def ValidateKey(cls, key):
+    for prefix in cls.VALID_PREFIXES:
+      if key.startswith(prefix):
+        return
+    raise errors.FormatError(u'Not a supported registry key prefix, got: {}'.format(key))
 
 
 class WindowsRegistryValueSourceType(SourceType):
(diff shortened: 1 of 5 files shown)


tests/source_type_test.py
```diff
@@ -42,7 +42,7 @@ class WindowsRegistryKeySourceTypeTest(unittest.TestCase):
 
   def testInitialize(self):
     """Tests the __init__ function."""
-    source_type.WindowsRegistryKeySourceType(keys=[u'test'])
+    source_type.WindowsRegistryKeySourceType(keys=[ur'HKEY_LOCAL_MACHINE\test'])
```
destijl: python3 doesn't like ur (see travis failure). Probably because all strings are unicode in 3. Use \\ here and below instead. Use tox to test it locally.

Lets also add a failure test where the keys is a string.

liamjm:
Fixed the unicode raw strings.

artifacts/source_type.py
```diff
@@ -165,6 +165,13 @@ class WindowsRegistryKeySourceType(SourceType):
 
   TYPE_INDICATOR = definitions.TYPE_INDICATOR_WINDOWS_REGISTRY_KEY
 
+  VALID_PREFIXES = [
+      r'HKEY_LOCAL_MACHINE',
+      r'HKEY_USERS',
+      r'HKEY_CLASSES_ROOT',
+      r'%%current_control_set%%',
```
joachimmetz: Why is current control set a valid prefix?

artifacts/source_type.py
```diff
@@ -178,9 +185,22 @@ def __init__(self, keys=None):
...
+
     super(WindowsRegistryKeySourceType, self).__init__()
     self.keys = keys
 
+  @classmethod
+  def ValidateKey(cls, key):
+    for prefix in cls.VALID_PREFIXES:
+      if key.startswith(prefix):
+        return
+    raise errors.FormatError(u'Not a supported registry key prefix, got: {}'.format(key))
```
joachimmetz: please have the style be consistent with the rest of the codebase, for now it uses {0:s} (or equiv)
{} => {0:s} (or equiv)

artifacts/source_type.py
```diff
@@ -165,6 +165,13 @@ class WindowsRegistryKeySourceType(SourceType):
...
 
+  VALID_PREFIXES = [
+      r'HKEY_LOCAL_MACHINE',
+      r'HKEY_USERS',
+      r'HKEY_CLASSES_ROOT',
+      r'%%current_control_set%%',
+  ]
+
   def __init__(self, keys=None):
     """Initializes the source type object.
```
joachimmetz: add a Raises section to the docstring

joachimmetz:
A couple of small issues, but no show stoppers.

There is some minor overlap with https://github.com/ForensicArtifacts/artifacts/pull/116 but let's get this PR in first and I'll patch up 116.

artifacts/source_type.py
destijl: For those following along, it is valid for now, but we should change it:
https://github.com/ForensicArtifacts/artifacts/issues/120

artifacts/source_type.py
liamjm: Done.

PR #141, opened 2015-11-25
Added Windows COM Registry keys.

joachimmetz:
[Review on Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/141)


destijl:
Looks like these might be missing CLSIDs...As discussed the empty value won't work in GRR.

joachimmetz:
I've added a * for the actual ids

destijl:
LGTM. I thought default value support for GRR was in, but it doesn't look like it and I couldn't find the change. Don't submit this until that's done.

destijl:
The default value support is in now:
https://github.com/google/grr/blob/master/lib/flows/general/collectors.py#L341

Issue #139, opened 2015-11-19
We should rename ARTIFACT to ARTIFACT_GROUP or similar


destijl:
Here's something I've seen a couple of times now, it's not obvious that the REGISTRY_VALUE is the artifact. People think they need to create a separate ARTIFACT object to hold it. In reality you only need to do this if you are grouping artifacts into a meta artifact.

ARTIFACT_GROUP
ARTIFACT_LIST
META_ARTIFACT

would probably be more intuitive names.

```
name: ConfickerIndicators
doc: Conficker Indicators.
sources:
- type: ARTIFACT
  attributes:
    names:
      - ConfickerWindowsRegKeys
supported_os: [Windows]
urls: ['https://www.iocbucket.com/iocs/e2903921f8b21d71dc88f34bee8b2f3491ea4656']
---
name: ConfickerWindowsRegKeys
doc: Windows Conficker IOCs.
sources:
- type: REGISTRY_VALUE
  attributes:
    key_value_pairs:
    - {key: 'HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\[RANDOM CHARACTERS]\"ImagePath"', value: '"%System%\svchost.exe -k netsvcs"'}
    - {key: 'HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\[WORM GENERATED SERVICE NAME]\"ImagePath"', value: '%SystemRoot%\system32\svchost.exe -k netsvcs'}
    - {key: 'HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\explorer\Advanced\Folder\Hidden\SHOWALL\"CheckedValue"', value: '0'}
    - {key: 'HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\Tcpip\Parameters\"TcpNumConnections"', value: '00FFFFFE'}
supported_os: [Windows]
urls: ['https://www.iocbucket.com/iocs/e2903921f8b21d71dc88f34bee8b2f3491ea4656']
```

destijl:
We should also add a test to reject artifact groups that only have one list item.

joachimmetz:
Both `ARTIFACT_LIST` and `ARTIFACT_GROUP` SGTM

joachimmetz:
Note to self update documentation

PR #143, opened 2015-11-25
Renamed ARTIFACT to ARTIFACT_GROUP #139

joachimmetz:
Changes for: https://github.com/ForensicArtifacts/artifacts/issues/139

[Review on Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/143)

### DIFF
--- a/artifacts/artifact.py
+++ b/artifacts/artifact.py
@@ -51,8 +51,8 @@ class ArtifactDefinition(object):
       raise errors.FormatError(u'Missing type indicator.')
 
     source_type_class = None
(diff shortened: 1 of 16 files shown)


destijl:
This LGTM but we need to make changes on the GRR side too. And we probably want to support both for a while, and warn people to change over their artifacts, so it might be a while before we submit this.

destijl:
GRR support is in for ARTIFACT_GROUP so this can go in after you merge.

joachimmetz:
Ack, I'll add a small documentation update to the CL as well.

joachimmetz:
Merged with upstream master, will merge PR once tests complete.

Issue #41, opened 2015-04-10
Improve validation/linting


joachimmetz:
Improve validation/linting to:

* detect multi line documentation string that does not have long docs with a white line after the first line. Also see [Style Guide](https://github.com/ForensicArtifacts/artifacts/blob/master/docs/style_guide.adoc#long-docs)

PR #145, opened 2015-11-27
Added check for duplicate Registry key paths #41

joachimmetz:
[Review on Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/145)

### DIFF
--- a/tools/validator.py
+++ b/tools/validator.py
@@ -31,6 +31,9 @@ class ArtifactDefinitionsValidator(object):
 
     Args:
       filename: the filename of the artifacts definition file.
+
+    Returns:
+      A boolean indicating the file contains valid artifacts definitions.
     """
     result = True
     artifact_reader = reader.YamlArtifactsReader()
--- a/tools/validator.py
+++ b/tools/validator.py
@@ -26,7 +26,7 @@ class ArtifactDefinitionsValidator(object):
     self._artifact_registry_key_paths = set()
     self._defined_artifact_names = set()
 
-  def _CheckDuplicateRegistryKeyPaths(
+  def _HasDuplicateRegistryKeyPaths(
       self, filename, artifact_definition, source):
     """Checks if Registry key paths are not already defined by other artifacts.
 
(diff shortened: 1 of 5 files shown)


tools/validator.py
```diff
@@ -44,10 +48,23 @@ def CheckFile(self, filename):
...
 
-        self.defined_artifact_names.add(artifact_definition.name)
+        self._defined_artifact_names.add(artifact_definition.name)
         for source in artifact_definition.sources:
-          if isinstance(source, source_type.ArtifactSourceType):
-            self.artifact_name_references.update(source.names)
+          # TODO: change to ARTIFACT_GROUP
+          if source.type_indicator == definitions.TYPE_INDICATOR_ARTIFACT:
+            self._artifact_name_references.update(source.names)
+
```
destijl: # Detect duplicate registry key paths

tools/validator.py
```diff
@@ -44,10 +48,23 @@ def CheckFile(self, filename):
...
+          elif source.type_indicator in (
+              definitions.TYPE_INDICATOR_WINDOWS_REGISTRY_KEY):
+            intersection = self._artifact_registry_key_paths.intersection(
+                set(source.keys))
+            if intersection:
+              duplicate_key_paths = u'\n'.join(intersection)
+              logging.warning((
```
destijl: I think we should set result=False here. Is there a case where we want to have an exact duplicate key? Seems like we should restructure artifacts if that is the case. We should warn if we're doing fuzzy matching, but exact dups should cause an error.

tools/validator.py
joachimmetz: I'll change this into a separate method so we don't need to add comments.

tools/validator.py
joachimmetz: Ack

joachimmetz:
Changed to _HasDuplicateRegistryKeyPaths

tools/validator.py
```diff
@@ -21,15 +21,47 @@ class ArtifactDefinitionsValidator(object):
...
     self._artifact_registry = registry.ArtifactDefinitionsRegistry()
-    self.defined_artifact_names = set()
-    self.artifact_name_references = set()
+    self._artifact_registry_key_paths = set()
+    self._defined_artifact_names = set()
+
+  def _HasDuplicateRegistryKeyPaths(
+      self, filename, artifact_definition, source):
+    """Checks if Registry key paths are not already defined by other artifacts.
+
```
destijl: Add comment: `Note this will only catch exact path duplicates.`

destijl:
LGTM, minor comment.

joachimmetz:
I've added it to the docstring instead

destijl:
Ah yep that's what I meant :)

PR #152, opened 2015-12-27
Changes to artifact names and definitions.

joachimmetz:
[Review on Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/152)


destijl:
This rename will break GRR interrogate:
https://github.com/google/grr/blob/master/config/artifacts.py#L27

I'm not sure it's worth the coordination effort to make this change (we'd have to do a new GRR server release at the same time). Or if we do it we should standardize on naming scheme everywhere (windows prefix etc.) and just do one big change next time we release the server.

joachimmetz:
> Or if we do it we should standardize on naming scheme everywhere (windows prefix etc.) and just do one big change next time we release the server.

This is likely the most time effective of the 2.

joachimmetz:
As discussed re-added the legacy definitions for backwards compatibility.

joachimmetz:
@destijl PTAL

definitions/windows_bootstrap.yaml
```diff
@@ -44,8 +102,8 @@ provides: [environ_programfiles]
 supported_os: [Windows]
 urls: ['http://environmentvariables.org/ProgramFiles']
 ---
-name: ProgramFilesx86
-doc: The %ProgramFiles (x86)% environment variable.
+name: WindowsEnvironmentVariableProgramFilesX86
```
destijl: I think we need to pick a consistent naming scheme. Either everything Windows starts with Windows, or nothing does.

joachimmetz:
Cleaned up Windows artifact naming
@destijl PTAL

destijl:
OK now that the GRR release is out I'm less worried about this breaking something, so lets get it merged with head and put it in. I'll look over it again once you have merged.

joachimmetz:
PTAL, please have a thorough look I needed to do a lot of merge fixes.

destijl:
This looks pretty good, although it is really hard to review so I wouldn't be surprised if I missed something. Couple of things remain: we should add code that requires artifacts in the windows.yaml file to start with Windows. Same for linux, which needs its own rename change. If our standards aren't enforced they won't be standards for long (we should also test for alphabetical order). This could be done in a follow-on cl if you like.

---

Reviewed 1 of 2 files at r1, 5 of 8 files at r2, 3 of 3 files at r3.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/forensicartifacts/artifacts/152#-:-KMBwUWS2dQtvV85doZW:1054980818)*

joachimmetz:
I opt a follow up CL, because implementing this could be tricky. As naive way could be everything in the windows.yaml file should start with Windows.

PR #160, opened 2016-02-10
Detect and handle bad keys in yaml definition.

destijl:
@joachimmetz

[Review on Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/160)

### DIFF
--- a/artifacts/artifact.py
+++ b/artifacts/artifact.py
@@ -82,6 +82,11 @@ class ArtifactDefinition(object):
       raise errors.FormatError(
           u'Unsupported type indicator: {0}.'.format(type_indicator))
 
-    source_object = source_type_class(**attributes)
+    try:
+      source_object = source_type_class(**attributes)
+    except (TypeError, AttributeError) as e:
+      raise errors.FormatError(
+          "Invalid artifact definition for {0}: {1}".format(self.name, e))
+
     self.sources.append(source_object)
     return source_object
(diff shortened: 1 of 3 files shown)


artifacts/artifact.py
```diff
@@ -82,6 +82,11 @@ def AppendSource(self, type_indicator, attributes):
       raise errors.FormatError(
           u'Unsupported type indicator: {0}.'.format(type_indicator))
 
-    source_object = source_type_class(**attributes)
+    try:
+      source_object = source_type_class(**attributes)
```
joachimmetz: Please make consistent with the rest of the code base e.g.
https://github.com/ForensicArtifacts/artifacts/blob/2fb0c26b355556e7a9b0db8af4b13d0bc0b56a26/artifacts/reader.py#L114

joachimmetz:
Style nit otherwise looks fine. When you fix the style I'll have another look.

artifacts/artifact.py
destijl: I'm not sure what you're asking for here?

joachimmetz: > raise errors.FormatError("Invalid artifact definition for %s: %s" % (self.name, e))

uses % instead of format which is used every else in the code base.

destijl: done. Also squashed the commits.

joachimmetz:
LGTM

PR #164, opened 2016-04-11
Add artifacts for linux syslogs and linux timezone file.

Onager:
This change is [Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/164)

### DIFF
--- a/definitions/linux.yaml
+++ b/definitions/linux.yaml
@@ -303,7 +303,7 @@ sources:
 labels: [Configuration Files, System]
 supported_os: [Linux]
 ---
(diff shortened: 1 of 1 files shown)


definitions/linux.yaml
```diff
@@ -303,6 +303,14 @@ sources:
 labels: [Configuration Files, System]
 supported_os: [Linux]
 ---
+name: LinuxSysLogs
```
joachimmetz: LinuxSysLogs => LinuxSysLogsFiles

PR #165, opened 2016-04-11
Add Linux log artifacts

berggren:
Add the following Linux log artifacts:

- /var/log/messages*
- /var/log/kern.log*
- /var/log/dpkg.log*
- /var/log/apt/history.log*

---
This change is [Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/165)

### DIFF
--- a/definitions/linux.yaml
+++ b/definitions/linux.yaml
@@ -83,6 +83,17 @@ urls:
   - http://manpages.ubuntu.com/manpages/saucy/man5/at.allow.5.html
   - http://manpages.ubuntu.com/manpages/precise/en/man1/crontab.1.html
 ---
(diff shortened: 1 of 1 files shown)


definitions/linux.yaml
```diff
@@ -483,3 +494,27 @@ urls:
...
+labels: [Logs]
+supported_os: [Linux]
+---
+name: LinuxKernelLogFiles
+doc: Linux kernel log files.
+sources:
+- type: FILE
+  attributes: {paths: ['/var/log/kern.log*']}
+labels: [Logs]
+supported_os: [Linux]
```
joachimmetz: + EOL

PR #168, opened 2016-05-25
Add Linux DSDTTable

raisfathin:
This change is [Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/168)

### DIFF
--- a/definitions/linux.yaml
+++ b/definitions/linux.yaml
@@ -534,3 +534,14 @@ sources:
   attributes: {paths: ['/var/log/kern.log*']}
 labels: [Logs]
 supported_os: [Linux]
+---
+name: DSDTTable
+doc: Linux file containing DSDT table.
+sources:
+- type: FILE
+  attributes:
+          paths: ['/sys/firmware/acpi/tables/DSDT']
+labels: [System]
+urls: ['https://www.kernel.org/doc/Documentation/acpi/initrd_table_override.txt']
+supported_os: [Linux]
+
--- a/definitions/linux.yaml
+++ b/definitions/linux.yaml
@@ -535,13 +535,11 @@ sources:
 labels: [Logs]
 supported_os: [Linux]
 ---
-name: DSDTTable
+name: LinuxDSDTTable
 doc: Linux file containing DSDT table.
 sources:
 - type: FILE
-  attributes:
-          paths: ['/sys/firmware/acpi/tables/DSDT']
+  attributes: {paths: ['/sys/firmware/acpi/tables/DSDT']}
 labels: [System]
 urls: ['https://www.kernel.org/doc/Documentation/acpi/initrd_table_override.txt']
 supported_os: [Linux]
-
--- a/definitions/linux.yaml
+++ b/definitions/linux.yaml
@@ -1,16 +1,5 @@
 # Linux specific artifacts.
 
-name: AllUsersShellHistory
-doc: Common shell history files for root and users.
-sources:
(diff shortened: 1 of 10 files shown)


definitions/linux.yaml
```diff
@@ -533,4 +533,15 @@ sources:
 - type: FILE
   attributes: {paths: ['/var/log/kern.log*']}
 labels: [Logs]
-supported_os: [Linux]
\ No newline at end of file
+supported_os: [Linux]
+---
+name: DSDTTable
```
joachimmetz: seeing this is linux specific artifact please prefix the name with Linux as: LinuxDSDTTable

definitions/linux.yaml
```diff
@@ -533,4 +533,15 @@ sources:
...
+doc: Linux file containing DSDT table.
+sources:
+- type: FILE
+  attributes:
+          paths: ['/sys/firmware/acpi/tables/DSDT']
+labels: [System]
+urls: ['https://www.kernel.org/doc/Documentation/acpi/initrd_table_override.txt']
+supported_os: [Linux]
+
```
joachimmetz: - white line

definitions/linux.yaml
```diff
@@ -533,4 +533,15 @@ sources:
...
-supported_os: [Linux]
\ No newline at end of file
+supported_os: [Linux]
+---
+name: DSDTTable
+doc: Linux file containing DSDT table.
+sources:
+- type: FILE
+  attributes:
+          paths: ['/sys/firmware/acpi/tables/DSDT']
```
joachimmetz: use same indentation as the other definitions

definitions/linux.yaml
```diff
@@ -533,4 +533,15 @@ sources:
...
+labels: [System]
+urls: ['https://www.kernel.org/doc/Documentation/acpi/initrd_table_override.txt']
+supported_os: [Linux]
```
joachimmetz: any specific version of Linux this was introduced?

joachimmetz:
Thanks for the PR, some minor style nits. Also see: https://github.com/ForensicArtifacts/artifacts/blob/master/docs/Artifacts%20definition%20format%20and%20style%20guide.asciidoc

raisfathin:
Fixed style issues, still searching for specific minimal Linux version which has this file.

raisfathin:
I found a [commit](https://github.com/torvalds/linux/commit/d4c5f047ae2a33296774e41abc2ac5c89283f736) which references creation of this file in v2.6.23.

Can you advise on how to write the conditions? I can't find any existing linux artifact that has `conditions` and I'm not quite sure about the value of `os_major_version` and `os_minor_version` on Linux.

destijl:
For GRR:
https://github.com/google/grr/blob/5a814fd020bc46beef79e88e2e4090634b68112c/grr/parsers/linux_release_parser.py#L182

It makes a lot less sense on linux than it does on windows. Lets just put it in without.

destijl:
You wouldn't know it but this is supposed to be alphabetical. Would you mind putting yours in order? If you can fix the others that are out of order also, that would be great.

raisfathin:
Alright.

I tried fixing the others, could you please check? I'm not sure if this could make the checking easier, but here is a little [script](https://gist.github.com/raisfathin/577555845be472bc99c0791a98f624c7) that I used to sort the artifacts.

destijl:
Thanks for going above and beyond. We should really add a test to check alphabetical order so it stays correct.

Issue #171, opened 2016-06-10
fix duplicate TYPE_INDICATOR_COMMAND


joachimmetz:
https://github.com/ForensicArtifacts/artifacts/blob/master/artifacts/artifact.py#L60

Originally reported by @jdavid

Issue #175, opened 2016-06-27
Add means to register custom artifact types


joachimmetz:
* [x] Add means to register custom artifact types.

PR #178, opened 2016-07-09
Artifact writer

pidydx:
These are changes I have made while working on building a ForensicArtifact server (more details soon).

I have heard there may be a lot of changes to this library coming soon so I wanted to get this up for initial review.  I still need to work out testing for the ArtifactWriter classes. 

This PR fixes: https://github.com/ForensicArtifacts/artifacts/issues/171

This PR is related to: https://github.com/ForensicArtifacts/artifacts/issues/175
I was actually going to implement the exact solution in the referenced PR and then found it was already waiting in a PR so I just left it at my minimal necessary changes for reference. Being able to add additional types would also be extremely useful, but at a minimum I need a list of valid types for data storage/searching in the artifact server.

I also made changes to ArtifactReader that allow overriding the definitions for similar reasons of being able to easily extend labels.

---
This change is [Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/178)


joachimmetz:
I'll split this PR in 2

I'll merge the fixes, but will leave some comments in this PR what I have changed

artifacts/definitions.py
```diff
@@ -12,6 +12,18 @@
 TYPE_INDICATOR_WINDOWS_REGISTRY_VALUE = 'REGISTRY_VALUE'
 TYPE_INDICATOR_WMI_QUERY = 'WMI'
 
+INDICATOR_TYPES = frozenset([
```
joachimmetz: => TYPE_INDICATORS

artifacts/reader.py
```diff
@@ -52,51 +67,76 @@ def ReadFileObject(self, file_object):
...
+
+    Raises:
+      FormatError: if the format of the artifact definition is not set
+                   or incorrect.
+    """
+
+
+class ArtifactsReader(BaseArtifactsReader):
+  def __init__(self):
+    self.supported_os = definitions.SUPPORTED_OS
```
joachimmetz: + call to super()

artifacts/reader.py
```diff
@@ -52,51 +67,76 @@ def ReadFileObject(self, file_object):
...
+      An artifact object (instance of ArtifactDefinition).
+
+    Raises:
+      FormatError: if the format of the artifact definition is not set
+                   or incorrect.
+    """
+
+
+class ArtifactsReader(BaseArtifactsReader):
+  def __init__(self):
```
joachimmetz: + docstring

artifacts/reader.py
```diff
@@ -52,51 +67,76 @@ def ReadFileObject(self, file_object):
...
+    Returns:
+      An artifact object (instance of ArtifactDefinition).
+
+    Raises:
+      FormatError: if the format of the artifact definition is not set
+                   or incorrect.
+    """
+
+
+class ArtifactsReader(BaseArtifactsReader):
```
joachimmetz: docstring and while line

tests/registry_test.py
```diff
@@ -27,7 +27,7 @@ def testArtifactDefinitionsRegistry(self):
     self.assertEqual(len(artifact_registry.GetDefinitions()), 7)
 
     artifact_definition = artifact_registry.GetDefinitionByName('EventLogs')
-    self.assertNotEquals(artifact_definition, None)
+    self.assertNotEqual(artifact_definition, None)
```
joachimmetz: => assertIsNotNone()

artifacts/source_type.py
```diff
@@ -40,6 +40,13 @@ def type_indicator(self):
           u'Invalid source type missing type indicator.')
     return self.TYPE_INDICATOR
 
+  @property
+  def attributes_definition(self):
+    """Dictionary representation of the source attributes
+
```
joachimmetz: no need for this white space.

artifacts/source_type.py
```diff
@@ -40,6 +40,13 @@ def type_indicator(self):
           u'Invalid source type missing type indicator.')
     return self.TYPE_INDICATOR
 
+  @property
+  def attributes_definition(self):
```
joachimmetz: Why is this an attribute?

artifacts/source_type.py
```diff
@@ -61,6 +68,14 @@ def __init__(self, names=None):
     super(ArtifactGroupSourceType, self).__init__()
     self.names = names
 
+  @property
+  def attributes_definition(self):
+    """Dictionary representation of the source attributes
+
+    """
+
+    return {'names': self.names}
```
joachimmetz: => u'names'

artifacts/reader.py
```diff
@@ -52,51 +67,76 @@ def ReadFileObject(self, file_object):
...
+  def __init__(self):
+    self.supported_os = definitions.SUPPORTED_OS
+    self.top_level_keys = definitions.TOP_LEVEL_KEYS
```
joachimmetz: why aren't these properties?

joachimmetz:
> Being able to add additional types would also be extremely useful, but at a minimum I need a list of valid types for data storage/searching in the artifact server.

Know that this will break portability of the artifact type. The reason for adding it will be that tools can implement their own for testing purposes. *It is highly recommended that the artifact type is added to the main project.*

artifacts/reader.py
pidydx: No reason.  I will change them to properties.

artifacts/source_type.py
pidydx: Also no reason.  I was actually going to implement it as a method at first and don't remember why I didn't.

pidydx:
ArtifactReader split out and ArtifactWriter updated

artifacts/writer.py
```diff
@@ -0,0 +1,132 @@
+# -*- coding: utf-8 -*-
+"""The artifact writer objects."""
+
+import abc
+import json
+import yaml
+
```
joachimmetz: style nit: + white line

artifacts/writer.py
```diff
@@ -0,0 +1,132 @@
...
+class BaseArtifactsWriter(object):
+  """Class that implements the artifacts reader interface."""
+
+  @abc.abstractmethod
+  def _ConvertArtifact(self, artifact):
```
joachimmetz: make this method name more explict e.g. CopyArtifactDefitionToDict also why is this method needed? Wouldn't it make more sense to add a CopyToDist() method to ArtifactDefinition ?

artifacts/writer.py
```diff
@@ -0,0 +1,132 @@
...
+    Args:
+      artifact: an instances of ArtifactDefinition
+
+    Yields:
+      dict representation of ArtifactDefinition instance
+    """
+
+  @abc.abstractmethod
+  def WriteArtifacts(self, artifacts, filename):
+    """Reads artifact definitions from a file.
```
joachimmetz: Reads => Writes

artifacts/writer.py
```diff
@@ -0,0 +1,132 @@
...
+
+    Args:
+      artifact: an instances of ArtifactDefinition
+
+    Yields:
+      dict representation of ArtifactDefinition instance
+    """
+
+  @abc.abstractmethod
+  def WriteArtifacts(self, artifacts, filename):
```
joachimmetz: make method name more explicit that you're writing to file (see reader)

artifacts/writer.py
```diff
@@ -0,0 +1,132 @@
...
+
+      filename: filename to write artifacts to
+
+    Yields:
+      writes artifacts to file
+    """
+
+  @abc.abstractmethod
+  def FormatArtifacts(self, artifacts):
+    """Formats artifact desired output format
```
joachimmetz: Unclear to me what this method is supposed to do

artifacts/writer.py
```diff
@@ -0,0 +1,132 @@
...
+
+    Args:
+      artifacts: an instance or list of ArtifactDefinition
+
+    Yields:
+      formatted string of artifact_definition
+    """
+
+
+class ArtifactWriter(BaseArtifactsWriter):
```
joachimmetz: + docstring

artifacts/writer.py
```diff
@@ -0,0 +1,132 @@
...
+      artifact_definition['supported_os'] = artifact.supported_os
+    if artifact.provides:
+      artifact_definition['provides'] = artifact.provides
+    if artifact.conditions:
+      artifact_definition['conditions'] = artifact.conditions
+    if artifact.urls:
+      artifact_definition['urls'] = artifact.urls
+    return artifact_definition
+
+  def WriteArtifacts(self, artifacts, filename):
```
joachimmetz: + docsring

artifacts/writer.py
```diff
@@ -0,0 +1,132 @@
...
+    with open(filename, 'w') as file_object:
+      file_object.write(self.FormatArtifacts(artifacts))
```
joachimmetz: FormatArtifacts is not implemented?

artifacts/writer.py
```diff
@@ -0,0 +1,132 @@
...
+    if artifact.provides:
+      artifact_definition['provides'] = artifact.provides
+    if artifact.conditions:
+      artifact_definition['conditions'] = artifact.conditions
+    if artifact.urls:
+      artifact_definition['urls'] = artifact.urls
+    return artifact_definition
+
+  def WriteArtifacts(self, artifacts, filename):
+    with open(filename, 'w') as file_object:
```
joachimmetz: what about binary formats?

artifacts/writer.py
```diff
@@ -0,0 +1,132 @@
...
+
+    Args:
+      artifacts: an instance or list of ArtifactDefinition
+
+    Returns:
+      formatted string of artifact_definition
+    """
+
+    if isinstance(artifacts, list):
+      artifact_definitions = [self._ConvertArtifact(artifact) for artifact in artifacts]
```
joachimmetz: why not have _ConvertArtifact correctly deal with a list?

artifacts/writer.py
```diff
@@ -0,0 +1,132 @@
...
+  def FormatArtifacts(self, artifacts):
+    """Converts artifacts to dict from ArtifactDefinition objects
+
+    Args:
+      artifacts: an instance or list of ArtifactDefinition
+
+    Returns:
+      formatted string of artifact_definition
+    """
+
```
joachimmetz: - white line

artifacts/writer.py
```diff
@@ -0,0 +1,132 @@
...
+    Args:
+      artifact: an instances of ArtifactDefinition
+
+    Yields:
+      dict representation of ArtifactDefinition instance
+    """
+    sources = []
+    for source in artifact.sources:
+      source_definition = {
+        'type': source.type_indicator,
```
joachimmetz: please use explicit unicode string u''

artifacts/writer.py
pidydx: Moving it to ArtifactDefinition as CopyToDict() is probably the better plan.  I will do that.

joachimmetz:
Can you elaborate what goal are you trying to accomplish with the writer?

pidydx:
Currently the main purpose I use the writer for is exporting artifacts from the Artificer server in YAML format, but the general idea is being able to take an ArtifactDefinition that has been read from some arbitrary source and validated and write it back in an arbitrary format (typically YAML for file and JSON for API)

pidydx:
Also using the Label/OS/etc tag filtering mentioned on Reader, but that is all still work in progress.

artifacts/writer.py
pidydx: I will have to look into this again with moving CopyToDict() to ArtifactDefinition.  The main headache here is actually the YAML side with dump vs dump_all and this was mostly for parity.  For JSON I could just as easily handle it returning a list of 1 dict instead of just the dict itself.

joachimmetz:
Let me give some thought on how to address your use case.

artifacts/writer.py
pidydx: Internally yes.  The Artificer server only uses FormatArtifacts to prepare YAML to serve rather than writing a file anywhere.

pidydx:
As an additional note, from one of the other discussions, having an ArtifactWriter could make it easier to deal sorting artifacts into their appropriate files, particularly as the collection grows.

pidydx:
I think I got everything corrected now.

destijl:
Reviewed 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 22 unresolved discussions.

---

*[artifacts/artifact.py, line 90 \[r3\]](https://reviewable.io:443/reviews/forensicartifacts/artifacts/178#-KN3sBJcNAXRE-FG9-Ma:-KN3l1t8XMxCzT-vNbcD:907733572) ([raw file](https://github.com/forensicartifacts/artifacts/blob/fe20a7a7993b50e03454a2d0b72a16246669f27b/artifacts/artifact.py#L90)):*
> ```Python
>     return source_object
> 
>   def CopyToDict(self):
> ```

Lets call it AsDict

I think this can be replaced by calling __dict__ or vars() on each source object and the top level object.

You also need to add a test to make sure this can convert every artifact in the repo.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/forensicartifacts/artifacts/178)*

pidydx:
Review status: 1 of 2 files reviewed at latest revision, 22 unresolved discussions.

---

*[artifacts/artifact.py, line 90 \[r3\]](https://reviewable.io:443/reviews/forensicartifacts/artifacts/178#-KN3sBJcNAXRE-FG9-Ma:-KNA2wC8D1Vdf8k1_532:967302890) ([raw file](https://github.com/forensicartifacts/artifacts/blob/fe20a7a7993b50e03454a2d0b72a16246669f27b/artifacts/artifact.py#L90)):*
<details><summary><i>Previously, destijl (Greg) wrote…</i></summary>
> Lets call it AsDict
> 
> I think this can be replaced by calling __dict__ or vars() on each source object and the top level object.
> 
> You also need to add a test to make sure this can convert every artifact in the repo.
</details>
I can change all of the CopyToDict to AsDict if that is what everyone likes.  I think AsDict might be a better name too.  

The main problem with the conversion is separators.  I don't want to have an artifact that doesn't define a separator come out defining a separator so I ignore the separator on each source CopyToDict if it is already the default value.

On the subject of tests, what all tests should be written and how would I test that each artifact converts correctly?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/forensicartifacts/artifacts/178)*

joachimmetz:
Review status: 1 of 2 files reviewed at latest revision, 22 unresolved discussions.

---

*[artifacts/artifact.py, line 90 \[r3\]](https://reviewable.io:443/reviews/forensicartifacts/artifacts/178#-KN3sBJcNAXRE-FG9-Ma:-KNAejOsH-3S9damH0YM:1028760549) ([raw file](https://github.com/forensicartifacts/artifacts/blob/fe20a7a7993b50e03454a2d0b72a16246669f27b/artifacts/artifact.py#L90)):*
<details><summary><i>Previously, pidydx (Sean Gillespie) wrote…</i></summary>
> I can change all of the CopyToDict to AsDict if that is what everyone likes.  I think AsDict might be a better name too.  
> 
> The main problem with the conversion is separators.  I don't want to have an artifact that doesn't define a separator come out defining a separator so I ignore the separator on each source CopyToDict if it is already the default value.
> 
> On the subject of tests, what all tests should be written and how would I test that each artifact converts correctly?
</details>
"CopyTo" is a commonly used convention in other code bases: https://msdn.microsoft.com/en-us/library/t69dktcd(v=vs.110).aspx

"As" should be fine as fine.

"dict or vars() on each source object and the top level object" 
I opt the keep the conversion explicit instead of relying on Python object internal behavior. Easier to maintain, less surprises if the internal behavior changes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/forensicartifacts/artifacts/178)*

destijl:
Review status: 1 of 2 files reviewed at latest revision, 22 unresolved discussions.

---

*[artifacts/artifact.py, line 90 \[r3\]](https://reviewable.io:443/reviews/forensicartifacts/artifacts/178#-KN3sBJcNAXRE-FG9-Ma:-KNDitU4ZZ-J04RSBX0_:-1467260146) ([raw file](https://github.com/forensicartifacts/artifacts/blob/fe20a7a7993b50e03454a2d0b72a16246669f27b/artifacts/artifact.py#L90)):*
<details><summary><i>Previously, joachimmetz (Joachim Metz) wrote…</i></summary>
> "CopyTo" is a commonly used convention in other code bases: https://msdn.microsoft.com/en-us/library/t69dktcd(v=vs.110).aspx
> 
> "As" should be fine as fine.
> 
> "dict or vars() on each source object and the top level object" 
> I opt the keep the conversion explicit instead of relying on Python object internal behavior. Easier to maintain, less surprises if the internal behavior changes.
> 
> 
</details>
Don't mind too much about the name,  we use asdict in GRR so it seems more familiar.

Two tests:

convert all artifacts in the repo to dict without raising
convert the test artifacts and compare to a golden dict (either defined in code or json)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/forensicartifacts/artifacts/178)*

pidydx:
This should be good for review, but it will need some fixes after https://github.com/ForensicArtifacts/artifacts/pull/180 gets merged.

tests/writer_test.py
```diff
@@ -59,9 +60,14 @@ def testYamlWriter(self):
     file_object = io.StringIO(initial_value=artifacts_yaml)
     converted_artifact_definitions = list(artifact_reader.ReadFileObject(
         file_object))
-    self.assertCountEqual(
-        [artifact.AsDict() for artifact in artifact_definitions],
-        [artifact.AsDict() for artifact in converted_artifact_definitions])
+    if sys.version_info >= (3, 0):
```
joachimmetz: What issue are you trying work around here? You can likely solve this in a different way.

artifacts/artifact.py
```diff
@@ -86,3 +86,40 @@ def AppendSource(self, type_indicator, attributes):
 
     self.sources.append(source_object)
     return source_object
+
+  def AsDict(self):
+    """Converts artifact to dict from ArtifactDefinition
```
joachimmetz: It does not convert since the original is preserved.

"""Represents an artifacts as a dictionary.

Returns:
   A dictionary containing the artifact attributes.
"""

artifacts/source_type.py
```diff
@@ -42,7 +42,7 @@ def type_indicator(self):
     return self.TYPE_INDICATOR
 
   @abc.abstractmethod
-  def CopyToDict(self):
+  def AsDict(self):
```
joachimmetz: Make the docstring consistent across the different AsDict methods.

artifacts/writer.py
```diff
@@ -0,0 +1,93 @@
+# -*- coding: utf-8 -*-
+"""The artifact writer objects."""
+
+import abc
+import json
+import yaml
+import sys
+from artifacts.artifact import ArtifactDefinition
```
joachimmetz: + file line before local imports

artifacts/writer.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class BaseArtifactsWriter(object):
+  """Class that implements the artifacts reader interface."""
```
joachimmetz: reader => writer

artifacts/writer.py
```diff
@@ -0,0 +1,93 @@
...
+
+  @abc.abstractmethod
+  def WriteArtifactsFile(self, artifacts, filename):
+    """Writes artifact definitions to a file.
+
+    Args:
+      artifacts: list of ArtifactDefinition objects to be written
+
```
joachimmetz: - white line

artifacts/writer.py
```diff
@@ -0,0 +1,93 @@
...
+  @abc.abstractmethod
+  def FormatArtifacts(self, artifacts):
+    """Formats artifacts to desired output format
+
+    Args:
+      artifacts: an instance or list of ArtifactDefinition
+
+    Returns:
+      formatted string of artifact_definition
```
joachimmetz: artifact_definition => artifact definition

artifacts/writer.py
```diff
@@ -0,0 +1,93 @@
...
+import yaml
+import sys
+from artifacts.artifact import ArtifactDefinition
+
+
+class BaseArtifactsWriter(object):
+  """Class that implements the artifacts reader interface."""
+
+  @abc.abstractmethod
+  def WriteArtifactsFile(self, artifacts, filename):
```
joachimmetz: Treat docstring as regular text and add punctuation

artifacts/writer.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class ArtifactWriter(BaseArtifactsWriter):
+
+  def WriteArtifactsFile(self, artifacts, filename):
+    """Writes artifact definitions to a file.
+
+    Args:
+      artifacts: list of ArtifactDefinition objects to be written
+
```
joachimmetz: - white line

artifacts/writer.py
```diff
@@ -0,0 +1,93 @@
...
+  def FormatArtifacts(self, artifacts):
+    """Converts artifacts to dict from ArtifactDefinition objects
+
+    Args:
+      artifacts: an instance or list of ArtifactDefinition
+
+    Returns:
+      formatted string of artifact_definition
+    """
+    if isinstance(artifacts, ArtifactDefinition):
```
joachimmetz: What if this fails?

tests/writer_test.py
```diff
@@ -0,0 +1,100 @@
...
+import unittest
+import yaml
+import json
+
+from artifacts import errors
+from artifacts import reader
+from artifacts import writer
+
+
+class ArtifactAsDictTest(unittest.TestCase):
```
joachimmetz: Why test the AsDict function here separately? Why not make it a test part of a Writer test case ?

artifacts/artifact.py
```diff
@@ -86,3 +86,40 @@ def AppendSource(self, type_indicator, attributes):
...
+
+    Returns:
+      dict representation of ArtifactDefinition instance
+    """
+    sources = []
+    for source in self.sources:
+      source_definition = {
+          'type': source.type_indicator,
```
joachimmetz: Make sure all your text string are unicode safe either with u'' or with importing the right future

artifacts/writer.py
```diff
@@ -0,0 +1,93 @@
...
+      artifacts = [artifacts]
+
+    artifact_definitions = [artifact.AsDict() for artifact in artifacts]
+    json_data = json.dumps(artifact_definitions)
+    if sys.version_info <= (3, 0):
```
joachimmetz: Solve this more elegant what if the json data is encoded?

artifacts/writer.py
```diff
@@ -0,0 +1,93 @@
...
+      json_data = unicode(json_data)
```
joachimmetz: unicode() is not portable and does not handle encodings, please don't use

tests/writer_test.py
```diff
@@ -0,0 +1,100 @@
...
+
+    last_artifact_definition = None
+    for artifact in artifact_definitions:
+      try:
+        artifact_definition = artifact.AsDict()
+      except errors.FormatError:
+        error_location = u'At start'
+        if last_artifact_definition:
+          error_location = u'After: {0}'.format(last_artifact_definition.name)
+        self.fail(u'{0} failed to convert to dict'.format(error_location))
```
joachimmetz: what are you testing here? Can you rewrite this test for it to become more clear from the code?

tests/writer_test.py
```diff
@@ -0,0 +1,100 @@
...
+    artifact_writer = writer.YamlArtifactsWriter()
+    test_file = os.path.join('test_data', 'definitions.yaml')
+
+    artifact_definitions = list(artifact_reader.ReadFile(test_file))
+    artifacts_yaml = artifact_writer.FormatArtifacts(artifact_definitions)
+
+    file_object = io.StringIO(initial_value=artifacts_yaml)
+    converted_artifact_definitions = list(artifact_reader.ReadFileObject(
+        file_object))
+    if sys.version_info >= (3, 0):
```
joachimmetz: Find an alternative solution for all the sys.version_info >= (3, 0) checks see comment below.

joachimmetz:
@pidydx left some comment, all the check for python 2 versus 3 looks very fragile at the moment. These can be solved in different ways by checking the data at hand not the python version.

tests/writer_test.py
pidydx: Python2 assertItemsEqual is renamed assertCountEqual in Python3.  I am open to better solutions as this was the simplest fix I could think of when fighting getting the unit tests to pass in 2 and 3.

artifacts/writer.py
pidydx: That is what I mean by an instance or list of ArtifactDefinition.  If you pass an instance it turns it into a list similar to how GRR handles similar cases.  I can change the docstring to be more clear or just force it to only accept lists.

tests/writer_test.py
pidydx: This tests that all artifacts in definitions can be converted to dict w/o raising.  All the other tests use the test definitions.

tests/writer_test.py
pidydx: I set this up to test that AsDict gets you the same dict as reading the artifact using yaml alone, then I test to make sure all the defined artifacts can be converted without raising.

The Writer tests are ensuring that, while using AsDict, they can convert to string output and read that string back in without losing anything.

tests/writer_test.py
pidydx: Related, should this just be except: instead of except errors.FormatError since the test is that it should be able to run AsDict on everything in definitions and not fail for any reason.

artifacts/writer.py
```diff
@@ -0,0 +1,102 @@
...
+    if isinstance(artifacts, ArtifactDefinition):
+      artifacts = [artifacts]
+
+    artifact_definitions = [artifact.AsDict() for artifact in artifacts]
+    json_data = json.dumps(artifact_definitions)
+    if isinstance(json_data, py2to3.BYTES_TYPE):
+      try:
+        return json_data.decode('ascii')
+      except UnicodeDecodeError:
+        pass
```
pidydx: @joachimmetz I just put pass here as a place holder for the moment.  What is the best thing to do in this case?  It seems like letting it just raise the UnicodeDecodeError rather than catching it makes the most sense.

artifacts/writer.py
joachimmetz: Please don't, stick with list if that is the required input.

tests/writer_test.py
joachimmetz: If a test raises if will fail no need to add the self.fail()

tests/writer_test.py
joachimmetz: TestCase is a collection of tests. at the moment these tests seem out of place. I opt to move them to the reader tests.

tests/writer_test.py
pidydx: I used self.fail as part of tracking the last successful artifact so the message points at exactly which artifact caused things to go wrong.

pidydx:
This should be mostly cleaned up now.  I need some feedback on the UnicodeDecodeError issue, but the rest of it should be correct now.

artifacts/writer.py
destijl: Add a comment about why this is necessary. I think you just need to make the error message as informative as possible (probably include the json data?) and raise as formaterror.

pidydx:
Everything here should be ready now.

tests/writer_test.py
```diff
@@ -0,0 +1,61 @@
...
+  """Class to test the YAML artifacts writer"""
+
+  def testYamlWriter(self):
+    """Tests the YamlArtifactsWriter FormatArtifacts method for loss during conversion."""
+    artifact_reader = reader.YamlArtifactsReader()
+    artifact_writer = writer.YamlArtifactsWriter()
+    test_file = os.path.join('test_data', 'definitions.yaml')
+
+    artifact_definitions = list(artifact_reader.ReadFile(test_file))
+    artifacts_yaml = artifact_writer.FormatArtifacts(artifact_definitions)
```
destijl: your ascii conversion problem is with stringIO right?

Maybe just use an actual temp file and that way you can also exercise the actual file writer code

with tempfile.TemporaryFile() as converted:
  write yaml to converted
  do checks

joachimmetz: FYI io.stringIO is unicode and io.bytesIO is binary.

pidydx: So I pushed a change to remove the py2to3 stuff in favor of StringIO/BytesIO if you would prefer to merge it that way, but I will probably switch this to actually doing the temp file if that is acceptable.

I only did all of this juggling because I assumed it would be better to not be writing/deleting temp files as part of unit tests, but if that isn't a concern then I can do it that way.

pidydx:
@destijl Good call on exercising the file write.  The juggling on writing changes introduced an error in py3.  Latest commit fixes that and uses tempfile for testing the Yamlwriter.  The JSON stuff doesn't need the io/file stuff and there is no JsonArtifactReader to really properly test reading from a json file anyways.

This should be ready now.  Let me know if there are any other necessary changes.

tests/writer_test.py
```diff
@@ -0,0 +1,58 @@
...
+class JsonArtifactsWriterTest(unittest.TestCase):
+  """Class to test the JSON artifacts writer"""
+
+  def testJsonWriter(self):
+    """Tests the JsonArtifactsWriter FormatArtifacts method for loss during conversion."""
+    artifact_reader = reader.YamlArtifactsReader()
+    artifact_writer = writer.JsonArtifactsWriter()
+    test_file = os.path.join('test_data', 'definitions.yaml')
+
+    artifact_definitions = list(artifact_reader.ReadFile(test_file))
```
destijl: Are you doing this to just avoid creating a temp file? It's confusing to see almost identical code tested quite differently. I don't think creating a tempfile is a problem given it's standard python, and it cleans up for you automatically.

pidydx: The reason for this is a lack of JsonArtifactReader.  I can write a json file to disk, but then the only code in the project that actually reads an artifact definition from a json file would be in the test so I just skipped the writing piece.

Because the yaml.safe_load_all() operates differently than json.loads() reading with 'rb' for json.loads() causes more py2/3 issues in writing a new JsonArtifactReader for a format that isn't supposed to really be on disk as JSON anyways.

destijl: Half-implementing JSON support is not ideal. I assume you do this because your GUI code will consume it? I think if you write a JSON writer, others users of the library will expect to find a reader.

pidydx: Modified to add a JsonArtifactReader

destijl: I think those two test methods are now identical: combine into one class with a helper method and just pass in the relevant reader/writer object.

pidydx: Updated to combined class.

destijl:
LGTM. I don't see anything outstanding from @joachimmetz so I'm going to merge.

PR #179, opened 2016-07-09
Added means to register custom artifact types #175

joachimmetz:
This implements the changes proposed in https://github.com/ForensicArtifacts/artifacts/pull/173 with interface functions.

---
This change is [Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/179)


.style.yapf
```diff
@@ -1,29 +1,31 @@
...
+SPLIT_BEFORE_NAMED_ASSIGNS = False
+SPLIT_PENALTY_AFTER_OPENING_BRACKET = 30
+SPLIT_PENALTY_AFTER_UNARY_OPERATOR = 10000
+SPLIT_PENALTY_BEFORE_IF_EXPR = 0
+SPLIT_PENALTY_BITWISE_OPERATOR = 300
+SPLIT_PENALTY_EXCESS_CHARACTER = 2600
+SPLIT_PENALTY_FOR_ADDED_LINE_SPLIT = 0
+SPLIT_PENALTY_IMPORT_NAMES = 0
+SPLIT_PENALTY_LOGICAL_OPERATOR = 300
+USE_TABS = False
```
destijl: lets just remove this completely and use --style chromium.

artifacts/registry.py
```diff
@@ -2,18 +2,62 @@
...
+
   def __init__(self):
     """Initializes the artifact definitions registry object."""
     super(ArtifactDefinitionsRegistry, self).__init__()
     self._artifact_definitions = {}
     self._artifact_name_references = set()
     self._defined_artifact_names = set()
 
+  @classmethod
+  def CreateSourceType(cls, type_indicator, attributes):
```
destijl: it seems odd that the registry is creating these objects. The registry should just return the correct class and the caller can create it. Call this "GetSourceTypeClass".

artifacts/registry.py
```diff
@@ -33,6 +77,25 @@ def DeregisterDefinition(self, artifact_definition):
 
     del self._artifact_definitions[artifact_definition_name]
 
+  @classmethod
+  def DeregisterSourceType(cls, source_type_class):
```
destijl: having a deregister without a register is confusing. Perhaps the better approach is to have

DEFAULT_TYPE_CLASSES

and in __init__ call a register method that registers them into the class dict.

artifacts/registry.py
```diff
@@ -87,6 +150,38 @@ def RegisterDefinition(self, artifact_definition):
...
+    """
+    if source_type_class.TYPE_INDICATOR in cls._source_type_classes:
+      raise KeyError(u'Source type already set for type: {0}.'.format(
+          source_type_class.TYPE_INDICATOR))
+
+    cls._source_type_classes[source_type_class.TYPE_INDICATOR] = (
+        source_type_class)
+
+  @classmethod
+  def RegisterSourceTypes(cls, source_type_classes):
```
destijl: Think it would be better to get rid of this and just call
```
for source_type_class in DEFAULT_SOURCE_TYPES:
 cls.register
```

in ```__init__```. Otherwise callers need to create the object and call this method with the private class variable?

joachimmetz:
Review status: 0 of 5 files reviewed at latest revision, 4 unresolved discussions.

---

*[.style.yapf, line 31 \[r1\]](https://reviewable.io:443/reviews/forensicartifacts/artifacts/179#-KMbmuuHmWZt87ZzzV0P-r1-31:-KN8VTPTEa30XaUnnvgY:-499146009) ([raw file](https://github.com/forensicartifacts/artifacts/blob/040cb1978f7b69e53e2a8653f84fd2b13397cf26/.style.yapf#L31)):*
<details><summary><i>Previously, destijl (Greg) wrote…</i></summary>
> lets just remove this completely and use --style chromium.
</details>
Done.

---

*[artifacts/registry.py, line 37 \[r1\]](https://reviewable.io:443/reviews/forensicartifacts/artifacts/179#-KMEUIDPe8N3OFLWxQVo-r1-37:-KN8W6laDCCw0IU81YPW:1606577899) ([raw file](https://github.com/forensicartifacts/artifacts/blob/040cb1978f7b69e53e2a8653f84fd2b13397cf26/artifacts/registry.py#L37)):*
<details><summary><i>Previously, destijl (Greg) wrote…</i></summary>
> it seems odd that the registry is creating these objects. The registry should just return the correct class and the caller can create it. Call this "GetSourceTypeClass".
</details>
The idea would be have this as a source type factory.

---

*[artifacts/registry.py, line 174 \[r1\]](https://reviewable.io:443/reviews/forensicartifacts/artifacts/179#-KMEUIDPe8N3OFLWxQVo-r1-174:-KN8W_07sxrap6Rx2MLN:-757244032) ([raw file](https://github.com/forensicartifacts/artifacts/blob/040cb1978f7b69e53e2a8653f84fd2b13397cf26/artifacts/registry.py#L174)):*
<details><summary><i>Previously, destijl (Greg) wrote…</i></summary>
> Think it would be better to get rid of this and just call
> 
> ```
> 
> for source_type_class in DEFAULT_SOURCE_TYPES:
> 
>  cls.register
> 
> ```
> 
> 
> 
> in ```__init__```. Otherwise callers need to create the object and call this method with the private class variable?
</details>
this is a class method so invocation would be
ArtifactDefinitionsRegistry.RegisterSourceTypes(...)

no need for instantiation of RegisterSourceTypes().

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/forensicartifacts/artifacts/179)*

joachimmetz:
@destijl not sure but it looks like reviewable is messing up the reply inside github.

destijl:
Review status: 0 of 5 files reviewed at latest revision, 6 unresolved discussions.

---

*[artifacts/registry.py, line 37 \[r1\]](https://reviewable.io:443/reviews/forensicartifacts/artifacts/179#-KMEUIDPe8N3OFLWxQVo-r1-37:-KN9gR_XmSWVlQ3GxilE:-1969619048) ([raw file](https://github.com/forensicartifacts/artifacts/blob/040cb1978f7b69e53e2a8653f84fd2b13397cf26/artifacts/registry.py#L37)):*
<details><summary><i>Previously, joachimmetz (Joachim Metz) wrote…</i></summary>
> The idea would be have this as a source type factory.
</details>
This just seems like a weird mix of roles that doesn't really save any code. It should just be a registry. As is it's a registry and a factory for only a subtype of artifact, but not artifacts themselves. Artifacts are created in a completely different module by a different class. 

---

*[artifacts/registry.py, line 81 \[r1\]](https://reviewable.io:443/reviews/forensicartifacts/artifacts/179#-KMEUIDPe8N3OFLWxQVo-r1-81:-KN9hRt0C9nXLQ7mGRlK:-2090980468) ([raw file](https://github.com/forensicartifacts/artifacts/blob/040cb1978f7b69e53e2a8653f84fd2b13397cf26/artifacts/registry.py#L81)):*
<details><summary><i>Previously, destijl (Greg) wrote…</i></summary>
> having a deregister without a register is confusing. Perhaps the better approach is to have
> 
> 
> 
> DEFAULT_TYPE_CLASSES
> 
> 
> 
> and in ```__init__``` call a register method that registers them into the class dict.
> 
> 
> 
> Oh my bad, there is a register...I'll keep reading.
</details>
we should put register and deregister next to each other. I think logical grouping is more important than strict alphabetical.

---

*[artifacts/registry.py, line 174 \[r1\]](https://reviewable.io:443/reviews/forensicartifacts/artifacts/179#-KMEUIDPe8N3OFLWxQVo-r1-174:-KN9kE-0bdyDMV5rxqt8:-1894884402) ([raw file](https://github.com/forensicartifacts/artifacts/blob/040cb1978f7b69e53e2a8653f84fd2b13397cf26/artifacts/registry.py#L174)):*
<details><summary><i>Previously, joachimmetz (Joachim Metz) wrote…</i></summary>
> this is a class method so invocation would be
> ArtifactDefinitionsRegistry.RegisterSourceTypes(...)
> 
> no need for instantiation of RegisterSourceTypes().
> 
</details>
I missed that these were class methods on the first pass. Since this class has a mix of methods that are very different: source type registration (class level) and artifact registration (object level) I think those should be grouped together rather than be sprinkled through the class in alphabetical order. It may even make more sense to split the source type registry into a different class.

---

*[tests/registry_test.py, line 46 \[r2\]](https://reviewable.io:443/reviews/forensicartifacts/artifacts/179#-KN9jP8Iad8bwbdWme8w:-KN9jP8Iad8bwbdWme8x:-2070569516) ([raw file](https://github.com/forensicartifacts/artifacts/blob/5e1984d9179e03b335f3d33d5d35f8a94e4070b4/tests/registry_test.py#L46)):*
> ```Python
>   """Tests for the artifact definitions registry."""
> 
>   # pylint: disable=protected-access
> ```

just add a GetSourceTypes method and call len on it rather than doing this.

---

*[tests/registry_test.py, line 116 \[r2\]](https://reviewable.io:443/reviews/forensicartifacts/artifacts/179#-KN9juPBcE1rX4Leeg6b:-KN9juPBcE1rX4Leeg6c:-1545973663) ([raw file](https://github.com/forensicartifacts/artifacts/blob/5e1984d9179e03b335f3d33d5d35f8a94e4070b4/tests/registry_test.py#L116)):*
> ```Python
> 
>     with self.assertRaises(KeyError):
>       registry.ArtifactDefinitionsRegistry.RegisterSourceType(TestSourceType)
> ```

break these into separate tests with descriptive names.

testRegisteringRaisesWhenAlreadyRegistered

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/forensicartifacts/artifacts/179)*

joachimmetz:
Review status: 0 of 5 files reviewed at latest revision, 6 unresolved discussions.

---

*[artifacts/registry.py, line 37 \[r1\]](https://reviewable.io:443/reviews/forensicartifacts/artifacts/179#-KMEUIDPe8N3OFLWxQVo-r1-37:-KNAiPQWdeS0YgImEFUb:1465237037) ([raw file](https://github.com/forensicartifacts/artifacts/blob/040cb1978f7b69e53e2a8653f84fd2b13397cf26/artifacts/registry.py#L37)):*
<details><summary><i>Previously, destijl (Greg) wrote…</i></summary>
> This just seems like a weird mix of roles that doesn't really save any code. It should just be a registry. As is it's a registry and a factory for only a subtype of artifact, but not artifacts themselves. Artifacts are created in a completely different module by a different class. 
</details>
ack moved to source type factory

---

*[artifacts/registry.py, line 81 \[r1\]](https://reviewable.io:443/reviews/forensicartifacts/artifacts/179#-KMEUIDPe8N3OFLWxQVo-r1-81:-KNAiD-eZ1DjND4_PyNq:160401764) ([raw file](https://github.com/forensicartifacts/artifacts/blob/040cb1978f7b69e53e2a8653f84fd2b13397cf26/artifacts/registry.py#L81)):*
<details><summary><i>Previously, destijl (Greg) wrote…</i></summary>
> we should put register and deregister next to each other. I think logical grouping is more important than strict alphabetical.
</details>
ack

---

*[artifacts/registry.py, line 174 \[r1\]](https://reviewable.io:443/reviews/forensicartifacts/artifacts/179#-KMEUIDPe8N3OFLWxQVo-r1-174:-KNAiW6KyRzijpmxJD48:160401764) ([raw file](https://github.com/forensicartifacts/artifacts/blob/040cb1978f7b69e53e2a8653f84fd2b13397cf26/artifacts/registry.py#L174)):*
<details><summary><i>Previously, destijl (Greg) wrote…</i></summary>
> I missed that these were class methods on the first pass. Since this class has a mix of methods that are very different: source type registration (class level) and artifact registration (object level) I think those should be grouped together rather than be sprinkled through the class in alphabetical order. It may even make more sense to split the source type registry into a different class.
</details>
ack

---

*[tests/registry_test.py, line 46 \[r2\]](https://reviewable.io:443/reviews/forensicartifacts/artifacts/179#-KN9jP8Iad8bwbdWme8w:-KNAiZUKt616jkpP5nt_:160401764) ([raw file](https://github.com/forensicartifacts/artifacts/blob/5e1984d9179e03b335f3d33d5d35f8a94e4070b4/tests/registry_test.py#L46)):*
<details><summary><i>Previously, destijl (Greg) wrote…</i></summary>
> just add a GetSourceTypes method and call len on it rather than doing this.
</details>
ack

---

*[tests/registry_test.py, line 116 \[r2\]](https://reviewable.io:443/reviews/forensicartifacts/artifacts/179#-KN9juPBcE1rX4Leeg6b:-KNAi_nHauKZYMdGthwe:160401764) ([raw file](https://github.com/forensicartifacts/artifacts/blob/5e1984d9179e03b335f3d33d5d35f8a94e4070b4/tests/registry_test.py#L116)):*
<details><summary><i>Previously, destijl (Greg) wrote…</i></summary>
> break these into separate tests with descriptive names.
> 
> testRegisteringRaisesWhenAlreadyRegistered
</details>
ack

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/forensicartifacts/artifacts/179)*

destijl:
:lgtm:

---

Reviewed 4 of 5 files at r1, 1 of 1 files at r2, 5 of 5 files at r3.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/forensicartifacts/artifacts/179#-:-KNDgkFBmB8ZzYDhGqaX:1417088721)*

PR #180, opened 2016-07-09
Fixes bugs in supported_os for Sources handling and refactors Artifac…

pidydx:
…tReader to allow for reading from other datasources or existing dicts

---
This change is [Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/180)

### DIFF
--- a/artifacts/reader.py
+++ b/artifacts/reader.py
@@ -4,16 +4,55 @@
 import abc
 import glob
 import os
+import yaml
 
 from artifacts import artifact
 from artifacts import definitions
 from artifacts import errors
 
-import yaml
 
-
-class ArtifactsReader(object):
+class BaseArtifactsReader(object):
   """Class that implements the artifacts reader interface."""
+
+  LABELS = None
+  SUPPORTED_OS = None
+
+  @property
+  def supported_os(self):
+    """The supported supported_os values.
+
+    Raises:
+      NotImplementedError: if the SUPPORTED_OS list is not defined.
+    """
+    if not self.SUPPORTED_OS:
+      raise NotImplementedError(u'Invalid supported SUPPORTED_OS list.')
+    return self.SUPPORTED_OS
+
+  @property
+  def labels(self):
+    """The supported label values.
+
+    Raises:
+      NotImplementedError: if the LABELS list is not defined.
+    """
+    if not self.LABELS:
+      raise NotImplementedError(u'Invalid supported LABELS list.')
+    return self.LABELS
+
+  @abc.abstractmethod
+  def ReadArtifactDefinition(self, artifact_definition):
+    """Reads an artifact definition.
+
+    Args:
+      artifact_definition: the artifact definition as a dict.
+
+    Returns:
+      An artifact object (instance of ArtifactDefinition).
+
+    Raises:
+      FormatError: if the format of the artifact definition is not set
+                   or incorrect.
+    """
 
   @abc.abstractmethod
   def ReadDirectory(self, path, extension=None):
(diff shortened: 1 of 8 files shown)


pidydx:
Main changes
BaseArtifactsReader - Abstract classes
ArtifactsReader - Defines all of the components required to turn a dict into an ArtifactDefinition
YamlArtifactsReader - Provides yaml to python dict reading
???ArtifactsReader - Allow for database or json or whatever to python dict for conversion into ArtifactDefinitions

Minor changes
Bug fix for dealing with Source supported_os keys.
Properties for labels and supported_os to allow for extending definitions

artifacts/definitions.py
```diff
@@ -40,7 +40,6 @@
     'IM': 'Instant Messaging / Chat applications artifacts.',
     'iOS' : 'Artifacts related to iOS devices connected to the system.',
     'History Files': 'History files artifacts e.g. .bash_history.',
-    'KnowledgeBase': 'Artifacts used in knowledge base generation.',
```
joachimmetz: this might be a GRR specific left over we cannot remove yet + @destijl for more insight

artifacts/reader.py
```diff
@@ -4,17 +4,58 @@
...
 
-class ArtifactsReader(object):
+class BaseArtifactsReader(object):
   """Class that implements the artifacts reader interface."""
 
+  LABELS = None
+  SUPPORTED_OS = None
+
+  @property
+  def supported_os(self):
```
joachimmetz: why define these as class constants?
why would the read need to define these properties?

let me if I understand what you want to achieve. Is your goal to allow to e.g. add custom labels?

artifacts/definitions.py
pidydx: This one is a duplicate key.

artifacts/reader.py
pidydx: Yes.  The purpose of the Artificer ForensicArtifacts server that I am building is to have an API to fetch/store/edit artifacts.  This tweak lets me add new OS/Types/Labels etc internally on the server and allow the reader to validate artifacts using them them as being correct.

Additionally on the Writer side it will eventually let me export artifacts from an artifact server using an arbitrary set of labels.  So if I  have labels I want for internal use, but aren't relevant to the broader set I can do proper validation using the additional labels, but when exporting to add back to a repo like this I can filter out unsupported values.

joachimmetz: Thx, your use case is suddenly much more clear

pidydx:
Does this need any more change or is this OK?

artifacts/reader.py
```diff
@@ -117,62 +165,53 @@ def _ReadArtifactDefinition(self, yaml_definition):
       if source_type:
         source_type.conditions = source.get(u'conditions', [])
         source_type.returned_types = source.get(u'returned_types', [])
-        self._ReadSupportedOS(yaml_definition, source_type, name)
+        self._ReadSupportedOS(source, source_type, name)
+        for supported_os in source_type.supported_os:
+          if supported_os not in artifact_object.supported_os:
```
destijl: this doesn't seem like a good idea. why not just add a validation step that requires the top level supported os to match any subkeys?

artifacts/definitions.py
```diff
@@ -40,7 +40,6 @@
     'IM': 'Instant Messaging / Chat applications artifacts.',
     'iOS': 'Artifacts related to iOS devices connected to the system.',
     'History Files': 'History files artifacts e.g. .bash_history.',
-    'KnowledgeBase': 'Artifacts used in knowledge base generation.',
```
destijl: there's only one of these now, do you still want to remove it? If so I'll have to go and see if the grr code uses it.

artifacts/reader.py
```diff
@@ -4,17 +4,56 @@
...
+
+  @property
+  def supported_os(self):
+    """The suppported_os values.
+
+    Raises:
+      NotImplementedError: if the type indicator is not defined.
+    """
+    if not self.SUPPORTED_OS:
+      raise NotImplementedError(u'Invalid source type missing type indicator.')
```
destijl: these error messages are inaccurate

artifacts/reader.py
pidydx: I was actually kind of curious about that myself.  The comments/documentation seem to indicate that that the intent is that supported_os, conditions, and returned_types are supposed to go away, but I would actually expect that supported_os and conditions should exist at the Source level.

Personally I would prefer to enforce that the top level supported_os match all of the defined sources, but I didn't want to do all that without understanding the intended direction.

destijl: we need both levels at the moment, they are both in use. I think where the top level doesn't match the lower levels is a bug in the definition we should be catching with validation.

artifacts/definitions.py
pidydx: This might not be correct anymore.  I removed this one because it was a duplicate key, but it appears that someone may have removed the other one as the duplicate key leaving this one in instead.

artifacts/reader.py
pidydx: Ok, I will just code this up to have a validation error if they don't match rather than just forcing the addition.

pidydx: Fixed

artifacts/definitions.py
pidydx: Fixed

artifacts/reader.py
pidydx: Fixed

pidydx:
Everything here should be corrected.  This PR fixes bugs in supported_os handling that need to be merged before ArtifactWriter UnitTests can pass properly. Getting this PR and ArtifactWriter in should cover all my immediate need use cases for my Artificer artifact server.

artifacts/reader.py
```diff
@@ -117,62 +166,57 @@ def _ReadArtifactDefinition(self, yaml_definition):
...
-    # TODO: check conditions.
-    artifact_definition.conditions = yaml_definition.get(u'conditions', [])
-    artifact_definition.provides = yaml_definition.get(u'provides', [])
-    self._ReadLabels(yaml_definition, artifact_definition, name)
-    self._ReadSupportedOS(yaml_definition, artifact_definition, name)
-    artifact_definition.urls = yaml_definition.get(u'urls', [])
+        self._ReadSupportedOS(source, source_type, name)
+        for supported_os in source_type.supported_os:
+          if supported_os not in artifact_object.supported_os:
+            source_supported_os.append(supported_os)
```
destijl: just raise here? you know the artifact is bad if this condition is met.

tests/reader_test.py
```diff
@@ -189,6 +189,23 @@ def testBadSupportedOS(self):
...
+- type: ARTIFACT_GROUP
+  attributes:
+    names:
+      - 'SystemEventLogEvtx'
+  supported_os: [Windows]
+labels: [Logs]
+""")
+
+    with self.assertRaises(errors.FormatError):
+      _ = list(artifact_reader.ReadFileObject(file_object))
```
destijl: you don't need the _ =

pidydx: I did that to match the other tests.

destijl:
LGTM. I don't see outstanding concerns from jbmetz so I will merge.

joachimmetz:
Sorry did not had the change yet to look at it things I have comments but I'll create CL to address them.

e.g. "Artifact object" is not correct.

joachimmetz:
Proposed clean up: https://github.com/ForensicArtifacts/artifacts/pull/192 issues addressed:

* We should not overload terms "Artifact object", Artifact definitions"
* Use of set.difference() IMO is more readable than using list comprehension
* yaml_file_object => just stick with file_object since that is the name used in the interface
* fixes linter errors

PR #184, opened 2016-07-11
Add GRR legacy names.

destijl:
Add back in some artifacts that were removed in:
 https://github.com/ForensicArtifacts/artifacts/pull/152

but are required for GRR interrogate to work.

---
This change is [Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/184)

### DIFF
--- a/tools/validator.py
+++ b/tools/validator.py
@@ -24,8 +24,8 @@ class ArtifactDefinitionsValidator(object):
     self._artifact_registry = registry.ArtifactDefinitionsRegistry()
     self._artifact_registry_key_paths = set()
 
-  def _HasDuplicateRegistryKeyPaths(
-      self, filename, artifact_definition, source):
+  def _HasDuplicateRegistryKeyPaths(self, filename, artifact_definition,
+                                    source):
     """Checks if Registry key paths are not already defined by other artifacts.
 
     Note that at the moment this function will only find exact duplicate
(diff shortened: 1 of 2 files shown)


destijl:
@joachimmetz

tools/validator.py
```diff
@@ -80,14 +80,18 @@ def CheckFile(self, filename):
         for source in artifact_definition.sources:
           if source.type_indicator in (
               definitions.TYPE_INDICATOR_WINDOWS_REGISTRY_KEY):
-            if self._HasDuplicateRegistryKeyPaths(
-                filename, artifact_definition, source):
+
+            # Exempt the legacy file from duplicate checking because it has
+            # duplicates intentionally.
+            if (filename != 'definitions/legacy.yaml' and
+                self._HasDuplicateRegistryKeyPaths(filename,
```
joachimmetz: why create all this white space?

tools/validator.py
```diff
@@ -80,14 +80,18 @@ def CheckFile(self, filename):
         for source in artifact_definition.sources:
           if source.type_indicator in (
               definitions.TYPE_INDICATOR_WINDOWS_REGISTRY_KEY):
-            if self._HasDuplicateRegistryKeyPaths(
-                filename, artifact_definition, source):
+
+            # Exempt the legacy file from duplicate checking because it has
+            # duplicates intentionally.
+            if (filename != 'definitions/legacy.yaml' and
```
joachimmetz: what about Windows? use os.path.join()

joachimmetz:
Please fix bug in this CL and I opt to revert the white space changes.

tools/validator.py
destijl: I use the auto python formatter since we started using it in grr. My editor is set to format on save, never have to fix a lint error again:
https://github.com/google/yapf

There's occasions when I would format differently, like this one. But the advantage of having it done automatically is greater I think.

tools/validator.py
destijl: I'll fix this in another change.

PR #185, opened 2016-07-12
Add missing GRR legacy names

destijl:
Also exempt legacy.yaml from duplicate checking.
Apply yapf python formatter.

---
This change is [Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/185)

### DIFF
--- a/tools/validator.py
+++ b/tools/validator.py
@@ -18,6 +18,8 @@ from artifacts import registry
 class ArtifactDefinitionsValidator(object):
   """Class to define an artifact definitions validator."""
 
+  LEGACY_PATH = os.path.join('definitions', 'legacy.yaml')
+
   def __init__(self):
     """Initializes the artifact definitions validator object."""
     super(ArtifactDefinitionsValidator, self).__init__()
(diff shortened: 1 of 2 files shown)


destijl:
@joachimmetz

tools/validator.py
```diff
@@ -24,8 +24,8 @@ def __init__(self):
     self._artifact_registry = registry.ArtifactDefinitionsRegistry()
     self._artifact_registry_key_paths = set()
 
-  def _HasDuplicateRegistryKeyPaths(
-      self, filename, artifact_definition, source):
+  def _HasDuplicateRegistryKeyPaths(self, filename, artifact_definition,
+                                    source):
```
joachimmetz: qq. why deviate from the style?

tools/validator.py
```diff
@@ -80,14 +80,18 @@ def CheckFile(self, filename):
         for source in artifact_definition.sources:
           if source.type_indicator in (
               definitions.TYPE_INDICATOR_WINDOWS_REGISTRY_KEY):
-            if self._HasDuplicateRegistryKeyPaths(
-                filename, artifact_definition, source):
+
+            # Exempt the legacy file from duplicate checking because it has
+            # duplicates intentionally.
+            if (filename != os.path.join('definitions', 'legacy.yaml') and
```
joachimmetz: I opt not to embed function calls in other logic too much makes debugging hard
legacy_yaml_filename = os.path.join('definitions', 'legacy.yaml')
maybe even make this a class constant?

if (filename != legacy_yaml_filename and self._HasDuplicateRegistryKeyPaths(
   ...

tools/validator.py
joachimmetz: per discussion: https://github.com/ForensicArtifacts/artifacts/issues/186

joachimmetz:
LGTM

PR #188, opened 2016-07-12
Add test that requires clean style to pass.

destijl:
- Run yapf over all code files
- Don't split before first arg to match Google style

---
This change is [Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/188)


.style.yapf
```diff
@@ -15,7 +15,7 @@ SPACES_BEFORE_COMMENT=2
 SPACE_BETWEEN_ENDING_COMMA_AND_CLOSING_BRACKET=False
 SPLIT_ARGUMENTS_WHEN_COMMA_TERMINATED=False
 SPLIT_BEFORE_BITWISE_OPERATOR=True
-SPLIT_BEFORE_FIRST_ARGUMENT=True
+SPLIT_BEFORE_FIRST_ARGUMENT=False
```
joachimmetz: Why? Turning this off creates harder to read constructions like:
+    artifact_definition = artifact.ArtifactDefinition(name,
 -        name, description=description)
+                                                      description=description)

where the treat this as a regular text (sentence) approach is much friendlier (less eye movement required).

artifacts/source_type.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-"""The source type objects.
+r"""The source type objects.
```
joachimmetz: Why not keep as is `r""""` is not anywhere else?

requirements.txt
```diff
@@ -1,2 +1,3 @@
 PyYAML==3.11
+yapf==0.10.0
```
joachimmetz: This is not a dependency! Move to setup_requires (or equiv)

https://pip.pypa.io/en/latest/reference/pip_install/#controlling-setup-requires

tests/style_test.py
```diff
@@ -0,0 +1,24 @@
+"""Enforce code style."""
+
+import subprocess
+import unittest
+
+from artifacts import errors
+
+
+class StyleTest(unittest.TestCase):
```
joachimmetz: + docstring so test information is shown during tests

tests/style_test.py
```diff
@@ -0,0 +1,24 @@
...
+
+  def testCodeStyle(self):
```
joachimmetz: + docstring so test information is shown during tests

artifacts/artifact.py
```diff
@@ -75,14 +75,14 @@ def AppendSource(self, type_indicator, attributes):
...
-          u'Unsupported type indicator: {0}.'.format(type_indicator))
+      raise errors.FormatError(u'Unsupported type indicator: {0}.'.format(
+          type_indicator))
 
     try:
       source_object = source_type_class(**attributes)
     except (TypeError, AttributeError) as e:
       raise errors.FormatError(
-          "Invalid artifact definition for {0}: {1}".format(self.name, e))
+          'Invalid artifact definition for {0}: {1}'.format(self.name, e))
```
joachimmetz: Please add: u'...'

.style.yapf
destijl: That's the problem with style, everyone thinks slightly differently, so I'd like to just adopt the google standard as closely as possible. I made this change to make it align better with google style so my editor set to enforce google style doesn't fight with yapf. I've asked the author if he can supply a yapf style file that mimicks internal style, we should adopt that if one exists. If not, this change at least fixed the most obvious difference.

artifacts/source_type.py
destijl: the google linter complains about it.

requirements.txt
destijl: I considered this, the problem is that most people who install this code will want to run the tests to validate any artifacts they write. Which will now break if this isn't installed. The way we solve this in grr is to have a separate package with test dependencies, but that is overkill in this case. I think adding a small dependency here is less bad than having tests break unexpectedly. I don't feel too strongly though, I can install it separately in the travis config if you like.

tests/style_test.py
destijl: done

tests/style_test.py
destijl: done

artifacts/source_type.py
joachimmetz: So this deviates from pylint

requirements.txt
joachimmetz: As discussed thought this applied to setup.py but only applies to travis configuration.

.style.yapf
joachimmetz: Google style guide allows for both. We are not following the [Google Python style guide](https://google.github.io/styleguide/pyguide.html) to the letter.

artifacts/definitions.py
```diff
@@ -34,9 +34,8 @@
     'Cloud Storage': 'Cloud storage artifacts.',
     'Configuration Files': 'Configuration files artifacts.',
     'Execution': 'Contain execution events.',
-    'ExternalAccount': (
-        'Information about any users\' account, e.g. username, '
-        'account ID, etc.'),
+    'ExternalAccount': ('Information about any user accounts e.g. username, '
+                        'account ID, etc.'),
     'External Media': 'Contain external media data or events e.g. USB drives.',
     'KnowledgeBase': 'Artifacts used in knowledge base generation.',
```
joachimmetz: Please remove duplicate key: KnowledgeBase

************* Module artifacts.definitions
W: 29, 9: Duplicate key 'KnowledgeBase' in dictionary (duplicate-key)

joachimmetz:
Per discussion let's set up the editor so that it uses the right versions of the tools.

LGTM

PR #192, opened 2016-07-26
Clean up of artifact reader after changes.

joachimmetz:
This change is [Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/192)

### DIFF
--- a/artifacts/reader.py
+++ b/artifacts/reader.py
@@ -4,6 +4,7 @@
 import abc
 import glob
 import os
+
 import yaml
 
 from artifacts import artifact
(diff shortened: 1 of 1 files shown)


artifacts/reader.py
```diff
@@ -167,59 +170,65 @@ def ReadArtifactDefinition(self, artifact_definition):
...
-    ]
+      raise errors.FormatError(u'Invalid supported_os type: {0}'.format(type(
+          supported_os)))
 
+    undefined_supported_os = set(supported_os).difference(self.supported_os)
     if undefined_supported_os:
-      error_string = (
-          u'Artifact definition: {0} supported operating system: {1} '
-          u'not defined.').format(name, u', '.join(undefined_supported_os))
+      if isinstance(definition_object, artifact.ArtifactDefinition):
```
destijl: This seems overly complicated. The reality is you already have the name of the artifact and the string of the undefined OS, which is plenty to track it down quickly. It doesn't really matter what level it is at inside the artifact.

destijl:
One minor comment, otherwise LGTM

artifacts/reader.py
joachimmetz: ack, remove the source type clause

Issue #25, opened 2015-03-05
document artifact definition naming convention


pidydx:
I have noticed there is a lot of variance between the names and labels for similar things.  What is the best way to align all of these things?

joachimmetz:
I'm working on cleaning up the current format https://github.com/ForensicArtifacts/artifacts/pull/11. Let us know what can be improved and I'll try to see if I can improve the "standard" and the documenation.

pidydx:
It is mostly things like how most of the OSX artifacts start with OSX, but the Linux ones rarely do.  For labels I think just some description around what belongs under each label.  Should OSXLaunchDaemons be System because of the system wide ones, or should it be User because of the user ones, or both, or something else?  That kind of thing.

joachimmetz:
Ack I'll add something to to style guide to make that more consistent and clean that up a bit afterwards.

joachimmetz:
Note to self update documentation to say:

* to keep the name space unique, prefix name of platform specific artifacts with the platform name
* update documentation
* update definitions

joachimmetz:
Dropping labels https://github.com/ForensicArtifacts/artifacts/issues/463

joachimmetz:
First stab at a definition naming convention https://artifacts.readthedocs.io/en/latest/sources/Format-specification.html#name

PR #194, opened 2016-07-29
Changes for artifact definition naming conventions #25

joachimmetz:
This change is [Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/194)


joachimmetz:
@destijl there is more to be cleaned up but if you can have a look at the initial changes to see if you agree with the direction in which it is heading.

destijl:
I don't think this is worth it. OSX prefix is fine, since they look to all have it, we should just leave it. Renaming is disruptive, there's probably artifacts grr depends on that will have to get added to legacy.

joachimmetz:
What about the other style notes?

destijl:
Sorry, didn't read that far down :) The doc changes LGTM.

joachimmetz:
Made a copy of artifact definition names hardcoded in GRR (https://github.com/google/grr/blob/master/grr/config/artifacts.py) into legacy

PR #197, opened 2016-08-29
Added registered type libraries definition.

joachimmetz:
This change is [Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/197)


joachimmetz:
@destijl PTAL

destijl:
Sorry, somehow I missed this. Empty value is what we decided on for default key value? We should put that in the style guide.

LGTM

PR #198, opened 2016-09-21
Updated artifact locations

superponible:
The Amcache.hve and RecentFileCache.bcf incorrectly have System32 in the path.  I haven't come across anything or seen a system where that's correct.

Added additional history locations for updated versions of Firefox and Opera.

---
This change is [Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/198)

### DIFF
--- a/definitions/webbrowser.yaml
+++ b/definitions/webbrowser.yaml
@@ -457,7 +457,9 @@ sources:
   supported_os: [Linux]
 - type: FILE
   attributes:
-    paths: ['%%users.appdata%%\Opera\Opera\global_history.dat']
+    paths:
+      - '%%users.appdata%%\Opera\Opera\global_history.dat'
+      - '%%users.appdata%%\Roaming\Opera Software\Opera Stable\History'
     separator: '\'
   supported_os: [Windows]
 supported_os: [Windows,Darwin,Linux]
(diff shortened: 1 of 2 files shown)


definitions/webbrowser.yaml
```diff
@@ -273,6 +273,7 @@ sources:
     paths:
       - '%%users.localappdata%%\Mozilla\Firefox\Profiles\*\places.sqlite'
       - '%%users.appdata%%\Mozilla\Firefox\Profiles\*\places.sqlite'
+      - '%%users.appdata%%\Roaming\Mozilla\Firefox\Profiles\*\places.sqlite'
```
joachimmetz: %APPDATA% should already contain Roaming
also see: https://en.wikipedia.org/wiki/Environment_variable

grrrrrrrrr: GRR documentation agrees, the roaming doesn't need to be there:
https://github.com/google/grr/blob/master/grr/proto/knowledge_base.proto#L136

superponible: In that case it looks like the Safari location is wrong since it contains Roaming there.  Do you need me to resubmit the PR without the FF change?

joachimmetz:
I've squashed merged this with some changes to save us from another round trip (back and forth on this PR).

PR #234, opened 2017-08-18
Split LSB and non-LSB release artifact definitions

joachimmetz:
This change is [Reviewable](https://reviewable.io/reviews/forensicartifacts/artifacts/234)


joachimmetz:
@grrrrrrrrr @demonchild2112 please have a look if this change affects GRR

PR #252, opened 2018-02-07
Fix usage of %%users.appdata%%

DeKe42:
When using the variable `%%users.appdata%%`, which is provided in the registry artifact `HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Explorer\Shell Folders`, the `Roaming` part is already included. This is consistent with other usages, for example with `users.localappdata`.

### DIFF
--- a/data/windows.yaml
+++ b/data/windows.yaml
@@ -1953,8 +1953,8 @@ sources:
 - type: FILE
   attributes:
     paths:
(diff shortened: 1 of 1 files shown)


joachimmetz:
Thx for the proposed fixe but %AppData% should include roaming. Also see: http://environmentvariables.org/AppData

joachimmetz:
Sorry not reading the changes correctly, will merge

PR #260, opened 2018-05-04
Bug fix for LinuxMessagesLogFiles definition

berggren:
The current definition points to /var/log/messages.log*, but the files are actually named /var/log/messages* (without .log)


berggren:
FYI: @joachimmetz @grrrrrrrrr

joachimmetz:
thx, LGMT

PR #261, opened 2018-05-07
Add MacOS bash session artifact

Karneades:
From the blog post https://www.swiftforensics.com/2018/05/bash-sessions-in-macos.html:

> While all versions of macOS have provided bash_history for users, since macOS 10.11 (El Capitan), we get even more information on terminal history through the bash sessions files. This is not a replacement for the old .bash_history file which is still there.
>
> The location you want to go to is  /Users/\<USER\>/.bash_sessions/*
>
> TERM_SESSION_ID.history    --> Contains session history
TERM_SESSION_ID.historynew --> Mostly blank/empty
TERM_SESSION_ID.session    --> Contains the last session resume date and time


joachimmetz:
thx, I've changed the formatting of urls into the short form, otherwise LGTM

PR #262, opened 2018-05-09
Added Windows system registry backup files artifact

Karneades:
Add artifact definition for primary files and transaction log files from RegBack Registry backup folder.


joachimmetz:
I observed that RegBack contains transaction files, but do you know if anything else than 0 byte files has been observed?

joachimmetz:
> Should I remove the artifact for RegBack transaction log artifact?

let's leave them, they add a purpose as an artifact definition, but I'll add a comment that so far these have been observed to be empty.

PR #264, opened 2018-05-17
Add INetCache folder to IE browser cache artifact

Karneades:


joachimmetz:
alas review further delayed due to limited time

I'll have a closer look as soon as time permits

Issue #265, opened 2018-05-24
Windows Path and File type sources missing path segment separator


jnettesheim:
Many artifact defintions in windows.yaml are missing a separator attribute for the FILE type, this means that a `/` is returned by default from artifacts when it should be a `\`.

joachimmetz:
Per chat I"ll add a validai check for this and correct the ones that are missing.

PR #268, opened 2018-06-11
Added check to validate Windows path separator #265

joachimmetz:
Added check to validate Windows path separator #265

### DIFF
--- a/tools/validator.py
+++ b/tools/validator.py
@@ -41,13 +41,13 @@ class ArtifactDefinitionsValidator(object):
     result = True
     key_path = key_path.upper()
 
-    if key_path.startswith(u'%%CURRENT_CONTROL_SET%%'):
+    if key_path.startswith('%%CURRENT_CONTROL_SET%%'):
       result = False
       logging.warning((
-          u'Artifact definition: {0:s} in file: {1:s} contains Windows '
-          u'Registry key path that starts with '
-          u'%%CURRENT_CONTROL_SET%%. Replace %%CURRENT_CONTROL_SET%% with '
-          u'HKEY_LOCAL_MACHINE\\System\\CurrentControlSet').format(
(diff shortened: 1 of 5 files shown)


tools/validator.py
```diff
@@ -100,12 +100,22 @@ def CheckFile(self, filename):
...
+              'Duplicate artifact definition: {0:s} in file: {1:s}'.format(
                   artifact_definition.name, filename))
           result = False
 
         for source in artifact_definition.sources:
-          if source.type_indicator == (
+          if source.type_indicator in (
+              definitions.TYPE_INDICATOR_FILE, definitions.TYPE_INDICATOR_PATH):
+            if source.supported_os == [definitions.SUPPORTED_OS_WINDOWS]:
+              if source.separator != '\\':
```
grrrrrrrrr: This check kind of implies that all artifacts that support Windows need to have the separator set to "\". If this is the case, why have the separator user defined at all?

joachimmetz: > why have the separator user defined at all?

There are various different types of path separators: https://en.wikipedia.org/wiki/Path_(computing) and artifacts are intended to self contain the necessary information in the definition. We could create a WINDOWS_PATH but IMHO that does not improve this. 

One additional thing the validator could check if the path contains more `\` than '/' for Windows. This change is a short term fix to ensure existing definition (for at least Windows) are explicit about their path segment separator.

grrrrrrrrr: Right I get the intention behind this. What I'm saying is that it will be confusing that when an artifact actually defines a path separator that is not '\', let's say that artifact wants to use '/' instead. In that case, the validation will complain with "Missing path separator...", which doesn't make sense.

joachimmetz: I'll add the additional check for the number of `\` versus '/' for now, but I somewhat agree with you on one side but on another side also would like this validated.

PR #280, opened 2018-08-22
Docker container logs & Worpdress config files

tomchop:
* WordPress configuration file
* Docker container log files

### DIFF
--- a/data/config_files.yaml
+++ b/data/config_files.yaml
@@ -23,3 +23,11 @@ sources:
   attributes: {paths: ['%%users.homedir%%/.ssh/config']}
 labels: [Configuration Files]
 supported_os: [Linux, Darwin]
+---
+name:  WordpressConfigFile
+doc: WordPress configuration file
+sources:
+- type: FILE
+  attributes: {paths: ['/**/wp-config.php']}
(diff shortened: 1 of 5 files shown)


data/config_files.yaml
```diff
@@ -23,3 +23,11 @@ sources:
   attributes: {paths: ['%%users.homedir%%/.ssh/config']}
 labels: [Configuration Files]
 supported_os: [Linux, Darwin]
+---
+name:  WordpressConfigFile
+doc: WordPress configuration file
+sources:
+- type: FILE
+  attributes: {paths: ['/**/wp-config.php']}
```
joachimmetz: Is there a way we could make this more specific, default wp installation locations? also know that this will only be 10 levels deep by default.

tomchop: Done, using the following three paths:

```
- '/wp/wp-config.php'
- '/var/www/wp-config.php'
- '/var/www/**/wp-config.php'
```

joachimmetz:
Merged manually in https://github.com/ForensicArtifacts/artifacts/commit/eef69bf52e41f485df05f3836da6d3c2bc189ceb

PR #289, opened 2018-09-04
Removing private/var/ duplicate artifacts

pstirparo:
removing/changing all private/var/ to var/ for consistency, as those are no new locations because var/ is a symlink to private/var/.

Also private/etc/ to etc/, as etc is symlink to private/etc tooo


joachimmetz:
Per conversation, please leave the alias paths seeing the definitions could be used in reverse, to map a path to an artifact definition as well.

PR #294, opened 2018-10-19
Several macOS new artifacts

pstirparo:
Added several new artifacts and corrected the path of the one related to Apple Mail application (new versions can have the path different from "V2", but still always V followed by one digit --> "V[0-9]")

### DIFF
--- a/data/macos.yaml
+++ b/data/macos.yaml
@@ -11,21 +11,6 @@ urls:
 - 'http://forensicswiki.org/wiki/Mac_OS_X'
 - 'http://forensicswiki.org/wiki/Mac_OS_X_10.9_-_Artifacts_Location#System_Logs'
 ---
-name: MacOSAppleUnifiedLog
-doc: Apple Unified Log
-sources:
-- type: FILE
-  attributes:
-    paths:
-    - '/var/db/diagnostics/*.tracev3'
-    - '/var/db/diagnostics/*/*.tracev3'
-    - '/var/db/uuidtext/*'
-labels: [System, Logs]
-supported_os: [Darwin]
-urls:
-- 'http://forensicswiki.org/wiki/Mac_OS_X'
-- 'https://github.com/mac4n6/Presentations/blob/master/Logs%20Unite!%20-%20Forensic%20Analysis%20of%20Apple%20Unified%20Logs/LogsUnite.pdf'
----
 name: MacOSApplications
 doc: Applications
 sources:
@@ -325,8 +310,7 @@ sources:
     - '/private/var/db/CoreDuet/Knowledge/knowledgeC.db'
 labels: [Users, Logs]
 supported_os: [Darwin]
-urls:
-- 'https://www.mac4n6.com/blog/2018/8/5/knowledge-is-power-using-the-knowledgecdb-database-on-macos-and-ios-to-determine-precise-user-and-application-usage'
+urls: ['https://www.mac4n6.com/blog/2018/8/5/knowledge-is-power-using-the-knowledgecdb-database-on-macos-and-ios-to-determine-precise-user-and-application-usage']
 ---
 name: MacOSLaunchAgentsPlistFiles
 doc: Launch Agents plist files
@@ -402,7 +386,7 @@ urls:
 - 'http://forensicswiki.org/wiki/Mac_OS_X_10.9_-_Artifacts_Location#System_Preferences'
 ---
 name: MacOSMailAccounts
-doc: Mail Accounts
+doc: Mail Accounts. Until now only V2, V3 and V5 have been observed.
 sources:
 - type: FILE
   attributes: {paths: ['%%users.homedir%%/Library/Mail/V[0-9]/MailData/Accounts.plist']}
@@ -413,7 +397,7 @@ urls:
 - 'http://forensicswiki.org/wiki/Mac_OS_X_10.9_-_Artifacts_Location#Mail'
 ---
 name: MacOSMailBackupTOC
-doc: Mail BackupTOC
+doc: Mail Backup Table of Content. Until now only V2, V3 and V5 have been observed.
 sources:
 - type: FILE
   attributes: {paths: ['%%users.homedir%%/Library/Mail/V[0-9]/MailData/BackupTOC.plist']}
@@ -424,7 +408,7 @@ urls:
 - 'http://forensicswiki.org/wiki/Mac_OS_X_10.9_-_Artifacts_Location#Mail'
 ---
 name: MacOSMailboxes
-doc: Mail Mailbox Directory
+doc: Mail Mailbox Directory. Until now only V2, V3 and V5 have been observed.
 sources:
 - type: FILE
   attributes: {paths: ['%%users.homedir%%/Library/Mail/V[0-9]/Mailboxes/*']}
@@ -446,7 +430,7 @@ urls:
 - 'http://forensicswiki.org/wiki/Mac_OS_X_10.9_-_Artifacts_Location#Mail'
 ---
 name: MacOSMailEnvelopIndex
-doc: Mail Envelope Index
+doc: Mail Envelope Index. Until now only V2, V3 and V5 have been observed.
 sources:
 - type: FILE
   attributes: {paths: ['%%users.homedir%%/Library/Mail/V[0-9]/MailData/Envelope Index']}
@@ -457,7 +441,7 @@ urls:
 - 'http://forensicswiki.org/wiki/Mac_OS_X_10.9_-_Artifacts_Location#Mail'
 ---
 name: MacOSMailIMAP
-doc: Mail IMAP Synched Mailboxes
+doc: Mail IMAP Synched Mailboxes. Until now only V2, V3 and V5 have been observed.
 sources:
 - type: FILE
   attributes: {paths: ['%%users.homedir%%/Library/Mail/V[0-9]/IMAP-*/*']}
@@ -468,7 +452,7 @@ urls:
 - 'http://forensicswiki.org/wiki/Mac_OS_X_10.9_-_Artifacts_Location#Mail'
 ---
 name: MacOSMailMainDirectory
-doc: Mail Main Folder
+doc: Mail Main Folder. Until now only V2, V3 and V5 have been observed.
 sources:
 - type: FILE
   attributes: {paths: ['%%users.homedir%%/Library/Mail/V[0-9]/*']}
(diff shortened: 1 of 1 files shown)


data/macos.yaml
```diff
@@ -11,6 +11,19 @@ urls:
 - 'http://forensicswiki.org/wiki/Mac_OS_X'
 - 'http://forensicswiki.org/wiki/Mac_OS_X_10.9_-_Artifacts_Location#System_Logs'
 ---
+name: MacOSAppleUnifiedLog
+doc: Apple Unified Log
+sources:
+- type: FILE
+  attributes:
+    paths:
+    - '/var/db/diagnostics/Persist/logdata.Persistent.YYYYMMDD THHMMSS.tracev3'
```
joachimmetz: I assume `YYYYMMDD THHMMSS` should be a wild card? also is the space in the string?

data/macos.yaml
```diff
@@ -369,7 +401,7 @@ name: MacOSMailAccounts
 doc: Mail Accounts
 sources:
 - type: FILE
-  attributes: {paths: ['%%users.homedir%%/Library/Mail/V2/MailData/Accounts.plist']}
+  attributes: {paths: ['%%users.homedir%%/Library/Mail/V[0-9]/MailData/Accounts.plist']}
```
joachimmetz: Have we seen all 0-9 versions? or is this 1-3 ?

data/macos.yaml
```diff
@@ -600,6 +658,14 @@ urls:
 - 'http://forensicswiki.org/wiki/Mac_OS_X'
 - 'http://forensicswiki.org/wiki/Mac_OS_X_10.9_-_Artifacts_Location#Skype'
 ---
+name: MacOSSkypeMainDirectory
```
joachimmetz: Seeing this is a cross platform application I opt to move this to applications definitions file

data/macos.yaml
```diff
@@ -600,6 +658,14 @@ urls:
...
+doc: Skype Directory
+sources:
+- type: DIRECTORY
+  attributes: {paths: ['%%users.homedir%%/Library/Application Support/Skype/*']}
+labels: [Users, Software, IM]
+supported_os: [Darwin]
+---
 name: MacOSSkypePreferences
```
joachimmetz: This one likely as well

data/macos.yaml
```diff
@@ -518,6 +550,26 @@ sources:
...
+- type: FILE
+  attributes:
+    paths:
+    - '/private/var/folders/[a-z][0-9]/*/0/com.apple.notificationcenter/db2/db'
+    - '/private/var/folders/[a-z][0-9]/*/0/com.apple.notificationcenter/db/db'
+    - '%%users.homedir%%/Library/Application Support/NotificationCenter/*.db'
+labels: [Users, Logs]
+supported_os: [Darwin]
+---
+name: MacOSPasswordHashes
```
joachimmetz: Doesn't this contain broader account information?

data/macos.yaml
```diff
@@ -292,6 +313,17 @@ urls:
 - 'http://forensicswiki.org/wiki/Mac_OS_X'
 - 'http://forensicswiki.org/wiki/Mac_OS_X_10.9_-_Artifacts_Location#Kernel_Extension'
 ---
+name: MacOSKnowledgeC
+doc: KnowledgeC User and Application usage database
```
joachimmetz: KnowledgeC ? what does this stand for?

Maybe add some urls:
https://www.mac4n6.com/blog/2018/8/5/knowledge-is-power-using-the-knowledgecdb-database-on-macos-and-ios-to-determine-precise-user-and-application-usage

data/macos.yaml
```diff
@@ -518,6 +550,26 @@ sources:
 labels: [System]
 supported_os: [Darwin]
 ---
+name: MacOSNotificationCenter
+doc: macOS NotificationCenter database
```
joachimmetz: Instead of `macOS` I opt to use the more generic `MacOS`, unless it is specific to versions of MacOS that are named `macOS` ;)

data/macos.yaml
pstirparo: I have seen V2, V3 and V5, I thought 0-9 was good to be on the safe side.

data/macos.yaml
pstirparo: Correct, I will change "YYYYMMDD THHMMSS" with "*".
The space was a typo

data/macos.yaml
pstirparo: Correct, it was a typo.

data/macos.yaml
joachimmetz: Know that "YYYYMMDDTHHMMSS" will result in an exact string match

not a glob "[0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9]T[0-9][0-9][0-9][0-9][0-9][0-9]"

data/macos.yaml
pstirparo: ok, I will remove it. Given that there were already Skype related artifacts I just added some new ones. I will start an im application specific yaml file with those in it.

data/macos.yaml
pstirparo: Fixed

data/macos.yaml
pstirparo: Fixed

data/macos.yaml
pstirparo: Fixed (see above)

data/macos.yaml
pstirparo: Indeed, it has other general account info like home directory, default shells, home dir path, original full user name, user profile picture, etc. Shall we use one of the following

1) name: MacOSUsersAccountsAndPassowrd
or
2) name: MacOSUsersAccountsAndPassowords
or just
3) name: MacOSUsersAccounts

with the "doc" being:
doc: Users Accounts Information and hashed Password plist files

data/macos.yaml
pstirparo: That DB is known as the "knowledgeC DB" that is why I left it as main name. Given it contains the user application usage, we may change it to either:
1)  "MacOSKnowledgeUserApplicationUsage", although it sounds too long
2) just "MacOSUserApplicationUsage", better although I would have liked to keep the KnowledgeC in the name, But probably this is more useful.

About the URL, yes it is in the URLs note of the mac4n6 artifacts spreadsheet, but I just didn't parse that field in the conversion. I will add it.

joachimmetz:
@pstirparo if ready for re-review, please re-assign me as reviewer or add a comment with PTAL

data/macos.yaml
pstirparo: As mentioned in my comments on the PR #298 , I'm referring specifically to the log files *.tracev3 and the resource files under uuidtext folder (this leaves out the log statistics file for example, which are in the same folder). would you prefer to keep them anyway?

data/macos.yaml
pstirparo: Actually this was already in the macos.yaml file as "MacOSUserPasswordHashesPlistFiles" same artifacts. I will remove it from this PR and address it in the next one, when cleaning the "plistfile" names

data/macos.yaml
pstirparo: ok, proper official name is "Duet Knowledge Base", changing the name to reflect this and moved "knowledgeC" and description to "doc:"

data/macos.yaml
```diff
@@ -11,6 +11,21 @@ urls:
...
 ---
+name: MacOSAppleUnifiedLog
+doc: Apple Unified Log
+sources:
+- type: FILE
+  attributes:
+    paths:
+    - '/var/db/diagnostics/*.tracev3'
+    - '/var/db/diagnostics/*/*.tracev3'
+    - '/var/db/uuidtext/*'
```
joachimmetz: `/var/db/uuidtext/*` will only get you the files in `/var/db/uuidtext/` you likely want `/var/db/uuidtext/*/*` as well?

data/macos.yaml
joachimmetz: thx SG

data/macos.yaml
```diff
@@ -292,6 +315,19 @@ urls:
...
+sources:
+- type: FILE
+  attributes:
+    paths:
+    - '%%users.homedir%%/Library/Application Support/Knowledge/knowledgeC.db'
+    - '/private/var/db/CoreDuet/Knowledge/knowledgeC.db'
+labels: [Users, Logs]
+supported_os: [Darwin]
+urls:
+- 'https://www.mac4n6.com/blog/2018/8/5/knowledge-is-power-using-the-knowledgecdb-database-on-macos-and-ios-to-determine-precise-user-and-application-usage'
```
joachimmetz: seeing this is a single url please change to `urls: ['https://...']`

data/macos.yaml
joachimmetz: ack, SG; maybe add this to the doc string that V2, V3 and V5 have been observed?

data/macos.yaml
```diff
@@ -11,6 +11,21 @@ urls:
 - 'http://forensicswiki.org/wiki/Mac_OS_X'
 - 'http://forensicswiki.org/wiki/Mac_OS_X_10.9_-_Artifacts_Location#System_Logs'
 ---
+name: MacOSAppleUnifiedLog
```
joachimmetz: please rename to "MacOSUnifiedLogging", since this is the name Apple is using.

https://developer.apple.com/videos/play/wwdc2016/721/
https://developer.apple.com/documentation/os/logging?language=occ

data/macos.yaml
```diff
@@ -11,6 +11,21 @@ urls:
...
+doc: Apple Unified Log
```
joachimmetz: Unified Log => Unified Logging / Unified Logging and Activity Tracing / Unified Logging and Tracing System

data/macos.yaml
```diff
@@ -380,7 +416,7 @@ name: MacOSMailBackupTOC
 doc: Mail BackupTOC
```
joachimmetz: I know this is not part of your changes but could you update this to "Mail Backup Table of Contents (TOC)", thx in advance.

data/macos.yaml
pstirparo: Done

data/macos.yaml
pstirparo: Yes, sorry it was a typo. First level is only directories.

data/macos.yaml
pstirparo: fixed

data/macos.yaml
pstirparo: fixed

data/macos.yaml
pstirparo: fixed

data/macos.yaml
pstirparo: changed to "Apple Unified Logging and Activity Tracing"

PR #295, opened 2018-10-22
Added artifact labels for plist and SQLite

pstirparo:
I think it would be beneficial to add three new labels for defining artifacts:
- DB and Plist are very useful in a macos (and mobile eventually) environment, where almost all data and configuration are stored in either format. 
- Persistence would be interesting so that we can mark known locations/artifacts of malware persistence mechanisms as such.

### DIFF
--- a/artifacts/definitions.py
+++ b/artifacts/definitions.py
@@ -34,11 +34,11 @@ LABELS = {
     'Mail': 'Mail client applications artifacts.',
     'Memory': 'Artifacts retrieved from memory.',
     'Network': 'Describe networking state.',
-    'Plist': 'Plist specific artifact.',
+    'Plist': 'Plist artifact.',
     'Processes': 'Describe running processes.',
     'Rekall': 'Artifacts using the Rekall memory forensics framework.',
     'Software': 'Installed software.',
-    'SQLiteDB': 'SQLite Database specific artifact.',
+    'SQLiteDB': 'SQLite database artifact.',
     'System': 'Core system artifacts.',
     'Users': 'Information about users.'
 }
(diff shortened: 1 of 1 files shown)


artifacts/definitions.py
```diff
@@ -19,6 +19,7 @@
     'Cloud': 'Cloud applications artifacts.',
     'Cloud Storage': 'Cloud storage artifacts.',
     'Configuration Files': 'Configuration files artifacts.',
+    'DB': 'Database specific artifact.',
```
joachimmetz: This is very generic. What use case did you have in mind? Maybe SQLiteDB ?

artifacts/definitions.py
```diff
@@ -34,6 +35,8 @@
     'Mail': 'Mail client applications artifacts.',
     'Memory': 'Artifacts retrieved from memory.',
     'Network': 'Describe networking state.',
+    'Persistence': 'Artifact location potentially used for malware persistence.',
```
joachimmetz: Maybe: `Artifact location potentially used for malware persistence.` => `Artifact known to be used for persistence of malware.`

artifacts/definitions.py
pstirparo: I thought about just DB because this would include all format, or else there should be one label for major DB types? I feel that could be too granular. 
My main use case comes from macOS artifacts, where most of the data from apps  are stored in sqlite db (and this applies to iOS and Android too). We could just have sqliteDB, sure.

artifacts/definitions.py
pstirparo: agree

joachimmetz:
> About persistence, I think would be interesting and useful to be able to pull all potential known persistent location for review in case of compromised machines.

Know that you could also do this with an artifact group.

PR #303, opened 2019-01-11
ntfs.yaml updates

macuisdein:
Added $LogFile as an artifact specific to NTFS.


joachimmetz:
@macuisdein thx for the proposed changes, sry for the slower response on my side.

I've made some minor tweaks but will merge the proposed changes.

PR #306, opened 2019-02-02
Small changes to OperaHistory definition

joachimmetz:


joachimmetz:
@grrrrrrrrr can you have a look if this impacts GRR

tools/validator.py
```diff
@@ -104,20 +184,18 @@ def CheckFile(self, filename):
...
 
+        artifact_definition_supports_windows = (
+            definitions.SUPPORTED_OS_WINDOWS in (
+                artifact_definition.supported_os))
+
         for source in artifact_definition.sources:
           if source.type_indicator in (
               definitions.TYPE_INDICATOR_FILE, definitions.TYPE_INDICATOR_PATH):
-            if definitions.SUPPORTED_OS_WINDOWS in source.supported_os:
+            if (artifact_definition_supports_windows or
```
grrrrrrrrr: Why not an and instead of a nested if?

tools/validator.py
```diff
@@ -104,20 +184,18 @@ def CheckFile(self, filename):
...
-                if (number_of_forward_slashes < number_of_backslashes and
-                    source.separator != '\\'):
-                  logging.warning((
-                      'Incorrect path separator: {0:s} in path: {1:s} defined '
-                      'by artifact definition: {2:s} in file: {3:s}').format(
-                          source.separator, path, artifact_definition.name,
-                          filename))
+                if not self._CheckWindowsPath(
+                    filename, artifact_definition, source, path):
                   result = False
```
grrrrrrrrr: A bit unrelated to this PR but all these could return straight away, wouldn't that be much easier to read?

tools/validator.py
```diff
@@ -52,6 +52,86 @@ def _CheckRegistryKeyPath(self, filename, artifact_definition, key_path):
...
+    if (number_of_forward_slashes < number_of_backslashes and
+        source.separator != '\\'):
+      logging.warning((
+          'Incorrect path separator: {0:s} in path: {1:s} defined '
+          'by artifact definition: {2:s} in file: {3:s}').format(
+              source.separator, path, artifact_definition.name,
+              filename))
+      result = False
+
+    if source.separator == '\\':
```
grrrrrrrrr: I'd go
if source.separator != '\\':
  \# Can't do any further validations
  return True

and move all the remaining code one level out.

tools/validator.py
```diff
@@ -52,6 +52,86 @@ def _CheckRegistryKeyPath(self, filename, artifact_definition, key_path):
...
+      path_lower = path.lower()
+      path_segments = path_lower.split(source.separator)
+
+      if path_segments[0].startswith('%%users.') and path_segments[0] not in (
```
grrrrrrrrr: Will this check work for this path I found in the same pr?
'%%users.localappdata%%\ConnectedDevicesPlatform\L.%%users.username%%\ActivitiesCache.db'

In GRR we use a regex to get the interpolations our, have a look at https://github.com/google/grr/blob/8123918157e0c53a2ba36d225cec542c6ee4524d/grr/core/grr_response_core/lib/artifact_utils.py#L75 Maybe you can reuse that code?

joachimmetz: Good point tracking this in https://github.com/ForensicArtifacts/artifacts/issues/310 for a follow up CL

tools/validator.py
joachimmetz: Mainly because I'm planning to add other checks e.g. https://github.com/ForensicArtifacts/artifacts/issues/309

tools/validator.py
joachimmetz: The idea is to keep checking, but return false at the end

PR #312, opened 2019-02-13
Expand glob braces

ant1:
This glob syntax is not documented and is not supported by plaso


joachimmetz:
Thx, I'll have a look as soon as time permits. Looks like the test are failing due to changes of the test environment separate from this PR.

PR #317, opened 2019-02-28
Change NTFSLogFile urls to list

cugu:
Urls should be a list.


joachimmetz:
Thx for the correction

Issue #309, opened 2019-02-04
Add validator check for Mac OS private paths


joachimmetz:
Per https://github.com/ForensicArtifacts/artifacts/pull/289 do not remove duplicates but add check to ensure duplicates are present.

* private/etc/ to etc/
* private/var/ to var/ (var/ is a symlink to private/var/)

Issue #274, opened 2018-07-16
Windows version conditions do not apply for Windows 10


DeKe42:
Some artifacts have conditions in the form of `os_major_version >= X AND os_minor_version >= Y`.
This fails starting with Windows 10, which can have version 10.0. A condition like `os_major_version >= 6 AND os_minor_version >= 1` will fail here although the artifact applies.

One example is
https://github.com/ForensicArtifacts/artifacts/blob/015b37543d97ad719f92d51b0a6cd548e4b9dab6/data/windows.yaml#L1335

I can think of multiple solutions to this:

- More complicated conditions like  `(os_major_version >= 6 AND os_minor_version >= 1) OR os_major version >= 7`
- A semantic version compare, something like `os_version >= "6.1"`

joachimmetz:
@DeKe42 thanks for the report. I'll consult with @grrrrrrrrr how GRR implements this. Moving to versions  like "6.1" looks like a better approach. Alternative is to drop the version all together.

joachimmetz:
Affected artifact definitions:

windows.yaml
WindowsActiveDesktop, WindowsAMCacheHveFile, WinAppXRT, WindowsEnvironmentVariableAppxProcess, WindowsEventLogApplication, WindowsEventLogSecurity, WindowsEventLogSystem, WindowsXMLEventLogTerminalServices, WindowsOpenSaveMRU, WindowsOpenSavePidlMRU, WindowsRecentFileCacheBCF, WindowsRunGrpConv, WindowsSetupApiLogs, 

wmi.yaml
WMIDrivers, WMIHotFixes, WMIInstalledSoftware, WMILoginUsers, WMIPhysicalMemory, WMIProcessList,

joachimmetz:
GRR CheckCondition applies an object filter to the knowledge base:
https://github.com/google/grr/blob/68a45fd446704285af25403d8dde4d3630524165/grr/core/grr_response_core/lib/artifact_utils.py#L222

joachimmetz:
GRR knowledge base gets filled with a separate value for `os_major_version` and `os_minor_version`, (likely via provides)

coperni:
I'm keen on taking this one up going for the latter solution with Semantic versioning. Assuming my changes get approved I will make a follow-up PR here to adjust some of these definitions.

joachimmetz:
[...]

PR #322, opened 2019-03-01
 Added validator check for Mac OS private paths #309

joachimmetz:


joachimmetz:
@pstirparo could you spot check this PR and see if this is inline with our conversation.

cugu:
Sorry for the interference, but what about 

``` yml
- type: FILE
  attributes: {paths: ['/private/etc/ssh/sshd_config']}
  supported_os: [Darwin]
- type: FILE
  attributes: {paths: ['/etc/ssh/sshd_config']}
  supported_os: [Darwin,Linux]
```

instead of

``` yml
- type: FILE
  attributes:
    paths:
    - '/etc/ssh/sshd_config'
    - '/private/etc/ssh/sshd_config'
  supported_os: [Darwin]
- type: FILE
  attributes: {paths: ['/etc/ssh/sshd_config']}
  supported_os: [Linux]
```

joachimmetz:
> Sorry for the interference, but what about

Interesting idea but IMHO a couple of downsides of that approach:

* slightly more complex conditional logic is necessary to validate
* easy to overlook the second source as one that applied to Darwin

For now I opt to keep it as is, and revisit this idea when I address the `supported_os` https://github.com/ForensicArtifacts/artifacts/issues/274

cugu:
Validation of duplicate paths #41 is more complicated this way.

joachimmetz:
> Validation of duplicate paths #41 is more complicated this way.

could you explain/elaborate why you think this is the case?

cugu:
With the current solution there are duplicates of file paths when ignoring the supported os. My solution prevents duplicate file paths.

joachimmetz:
> My solution prevents duplicate file paths.

How? I'm missing the necessary context.

Your suggested approach reduces the **need** for duplication in a single artifact definition/
However duplication can still exist across artifact definitions, which is what hinted at in #41.

cugu:
A simple validator for duplicate file paths will find the duplicate `/etc/ssh/sshd_config` in 

``` yaml
- type: FILE
  attributes:
    paths:
    - '/etc/ssh/sshd_config'
    - '/private/etc/ssh/sshd_config'
  supported_os: [Darwin]
- type: FILE
  attributes: {paths: ['/etc/ssh/sshd_config']}
  supported_os: [Linux]
```

but not in 
```yaml
- type: FILE
  attributes:
    paths:
    - '/etc/ssh/sshd_config'
    - '/private/etc/ssh/sshd_config'
  supported_os: [Darwin]
- type: FILE
  attributes: {paths: ['/etc/ssh/sshd_config']}
  supported_os: [Linux]
```

Anyway it does not matter to much, I just have to make my duplication checker aware of different supported_os.

joachimmetz:
> Anyway it does not matter to much, I just have to make my duplication checker aware of different supported_os.

Note that supported_os is on the list to be changed, I opt not to invest too much effort here yet.

PR #333, opened 2019-04-25
[WIP] Fix several existing artifacts, add in new ones

recvfrom:
One of the registry keys related to the Logon Script environment persistence mechanism incorrectly listed 'UserMprLogonScript' as the environment variable name when it should be 'UserInitMprLogonScript' (per http://www.hexacorn.com/blog/2014/11/14/beyond-good-ol-run-key-part-18/).

Also, two of the registry keys had the invisible soft hyphen character inserted.  I'm not sure if this affects things when these data files are used, but it's probably best to remove them. `git diff` from the console shows these changes better than the GitHub UI:
```diff
-      - '%%environ_systemroot%%\system32\Windows<U+00AD>PowerShell\v1.0\profile.ps1'
-      - '%%environ_systemroot%%\system32\Windows<U+00AD>PowerShell\v1.0\Microsoft.PowerShell_profile.ps1'
+      - '%%environ_systemroot%%\system32\WindowsPowerShell\v1.0\profile.ps1'
+      - '%%environ_systemroot%%\system32\WindowsPowerShell\v1.0\Microsoft.PowerShell_profile.ps1'
```

### DIFF
--- a/data/windows.yaml
+++ b/data/windows.yaml
@@ -2411,24 +2411,29 @@ doc: Used to filter TCP/IP traffic through WinSock2.
 sources:
 - type: REGISTRY_KEY
   attributes:
-    keys: ['HKEY_LOCAL_MACHINE\System\CurrentControlSet\Services\WinSock2\Parameters\Protocol_Catalog9\Catalog_Entries\*']
+    keys:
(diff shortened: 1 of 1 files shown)


joachimmetz:
Thx for the PR, I'll have a look as soon as time permits

recvfrom:
From doing some research, I found more artifacts that could be added in.  There are a few others I plan to add in next week, but I figured I'd push what I had thus far so you all can tell me if I'm way off base on these.

PR #347, opened 2019-06-11
Adjusted typo of CurrentControlSet

marcurdy:
Small change.

### DIFF
--- a/data/windows.yaml
+++ b/data/windows.yaml
@@ -3032,7 +3032,7 @@ sources:
     key_value_pairs:
       - {key: 'HKEY_LOCAL_MACHINE\System\CurrentControlSet\Services\Tcpip\Parameters', value: 'NameServer'}
       - {key: 'HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\Tcpip\Parameters\Interfaces\*', value: 'NameServer'}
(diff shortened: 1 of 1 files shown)


joachimmetz:
Thx, for the corrections

marcurdy:
YW. I've created a script to use your yamls as input and create plaso tag
files.

joachimmetz:
Ack, know that plaso directly supports artifacts as collection filters as well

marcurdy:
I only learned this 20 minutes ago. Thank you for your huge contributions!

One question on the windows yaml.
WindowsCOMLocalServers contains different description styles. One places
the LocalServer as a value. The second puts it alongside the key and leaves
the value empty. Is this intentional?

            "key": "HKEY_LOCAL_MACHINE\\Software\\Classes\\CLSID\\*",
            "value": "LocalServer"
          }, {
            "key":
"HKEY_LOCAL_MACHINE\\Software\\Classes\\CLSID\\*\\LocalServer32",
            "value": ""

WindowsCOMInprocServers puts all values in the key leaving value along.

Cheers

joachimmetz:
The first means key with value LocalServer, the second means default value of key. I'll double check but the second could be an error. For GRR historically this does not make a difference because the value is appended to the key path (which can cause issues with e.g. bagmru paths where subkey and value names are the same).

PR #349, opened 2019-06-13
Added WindowsUserJumpLists artifact definitions

ant1:


data/windows.yaml
```diff
@@ -2768,6 +2768,20 @@ sources:
 labels: [Users]
 supported_os: [Windows]
 ---
+name: WindowsUserJumpLists
```
joachimmetz: Please make this a group and split the different jumplist types into separate definitions 

This allows additions of other (lesser known) jump list formats, such as Explorer ProgramsCache Registry values. Also see: https://github.com/libyal/dtformats/blob/master/documentation/Jump%20lists%20format.asciidoc

joachimmetz:
@ant1 thx for the addition, though I have a request to change the PR a bit.

data/windows.yaml
ant1: This should be fixed

joachimmetz:
@ant1 excellent, thx. I'll have a closer look later today.

PR #350, opened 2019-06-14
Changes to Jump Lists artifact definitions

joachimmetz:


joachimmetz:
@ant1 thx for double checking.

PR #356, opened 2019-06-27
Add/Update multiple Windows artifact definitions

recvfrom:
Add WindowsApplicationShimFiles
Add WindowsApplicationShimRegistryKeys
Add WindowsCOMProperties
Add WindowsBootConfigurationSettings
Add WindowsUpdateSettings
Update WindowsExplorerSettings
Update WindowsSystemRestoreSettings
Update WindowsUserAccountControlSettings

### DIFF
--- a/data/windows.yaml
+++ b/data/windows.yaml
@@ -109,12 +109,15 @@ urls:
 - 'http://www.hexacorn.com/blog/2013/01/19/beyond-good-ol-run-key-part-3/'
 - 'https://msdn.microsoft.com/en-us/library/windows/desktop/ee872121(v=vs.85).aspx'
 ---
-name: WindowsApplicationShimFiles
+name: WindowsApplicationCompatibilityInstalledShimDatabases
 doc: |
-  Windows Application Shim Files
+  Windows Application Compatibility Installed Shim Databases.
 
-  Malicious shim database files can be used to bypass UAC, inject DLLs into
-  legitimate processes, and modify anti-virus software to avoid detection.
+  Malicious shim database file entries can be used to bypass UAC, inject DLLs
+  into legitimate processes, and modify anti-virus software to avoid detection.
+  See the description under the
+  WindowsApplicationCompatibilityShimDatabaseMappings artifact for a more
+  in-depth overview of Windows application compatibility shims.
 
   drvmain.sdb, frxmain.sdb, msimain.sdb, pcamain.sdb, and sysmain.sdb are
   shim database files that ship with Windows, and contain many predefined
@@ -122,7 +125,8 @@ doc: |
   signed, so malware can insert shim entries into them.
 
   The other paths point to the locations where SDB files are installed
-  when the sdbinst.exe utility is used to install them.
+  when the sdbinst.exe utility is used to install them. Note, however, that
+  shim database files can also exist elsewhere in the file system.
 sources:
 - type: FILE
   attributes:
@@ -145,12 +149,29 @@ urls:
 - 'http://files.brucon.org/2015/Tomczak_and_Ballenthin_Shims_for_the_Win.pdf'
 - 'https://www.blackhat.com/docs/eu-15/materials/eu-15-Pierce-Defending-Against-Malicious-Application-Compatibility-Shims-wp.pdf'
 ---
-name: WindowsApplicationShimRegistryKeys
+name: WindowsApplicationCompatibilityShimDatabaseMappings
 doc: |
-  Windows Application Shims Registry Keys
+  Windows Application Compatibility Shim Database Mappings.
 
-  These can be used by malware to bypass UAC, inject DLLs into legitimate
-  processes, and modify anti-virus software to avoid detection.
+  Windows application shims provide a way for the operating system to
+  apply patches to executables before they are run, ultimately providing
+  a lightweight mechanism for applying hot fixes and making modifications to
+  ensure compatibility across the various versions of Windows. This
+  functionality can also be leveraged maliciously to change how certain
+  programs operate, or to provide capabilities to malware, such as the
+  ability to bypass UAC, gain persistence by injecting loading into legitimate
+  processes, or avoid detection by disabling anti-virus software.
+
+  Windows allows for custom application shims to be installed via the
+  sdbinst.exe application, which stores the mappings between the the shim
+  database files and the programs that they apply to in the registry. An
+  example registry layout for an application shim applied to the 'notepad.exe'
+  application would look like (all keys relative to
+  HKEY_LOCAL_MACHINE\Software\Microsoft\Windows NT\CurrentVersion\):
+  * AppCompatFlags\Custom\notepad.exe\{00000000-1111-2222-3333-444444444444}.sdb = 0
+  * AppCompatFlags\InstalledSDB\{00000000-1111-2222-3333-444444444444}\DatabasePath = 
+      "C:\Windows\AppPatch\Custom\{00000000-1111-2222-3333-444444444444}.sdb"
+
 sources:
 - type: REGISTRY_VALUE
   attributes:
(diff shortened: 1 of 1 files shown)


data/windows.yaml
```diff
@@ -109,6 +109,60 @@ urls:
 - 'http://www.hexacorn.com/blog/2013/01/19/beyond-good-ol-run-key-part-3/'
 - 'https://msdn.microsoft.com/en-us/library/windows/desktop/ee872121(v=vs.85).aspx'
 ---
+name: WindowsApplicationShimFiles
```
joachimmetz: please change to:
ShimFiles => ShimDatabaseFiles

data/windows.yaml
```diff
@@ -109,6 +109,60 @@ urls:
...
+supported_os: [Windows]
+urls:
+- 'https://attack.mitre.org/techniques/T1138/'
+- 'https://countercept.com/blog/hunting-for-application-shim-databases/'
+- 'http://files.brucon.org/2015/Tomczak_and_Ballenthin_Shims_for_the_Win.pdf'
+- 'https://www.blackhat.com/docs/eu-15/materials/eu-15-Pierce-Defending-Against-Malicious-Application-Compatibility-Shims-wp.pdf'
+---
+name: WindowsApplicationShimRegistryKeys
+doc: |
+  Windows Application Shims Registry Keys
```
joachimmetz: Registry => Windows Registry

data/windows.yaml
```diff
@@ -109,6 +109,60 @@ urls:
...
+
+  These can be used by malware to bypass UAC, inject DLLs into legitimate
```
joachimmetz: Does comment applies to the values or to the shims?

data/windows.yaml
joachimmetz: and Keys => Values

joachimmetz: maybe rename this key to closer reflect its purpose? isn't it used to store the location of the shim database files?

data/windows.yaml
```diff
@@ -331,6 +385,39 @@ urls:
...
+    key_value_pairs:
+    - {key: 'HKEY_LOCAL_MACHINE\Software\Classes\CLSID\*\ShellFolder', value: 'Attributes'}
+    - {key: 'HKEY_USERS\%%users.sid%%\Software\Classes\CLSID\*\ShellFolder', value: 'Attributes'}
+    - {key: 'HKEY_LOCAL_MACHINE\Software\Classes\Wow6432Node\CLSID\*\ShellFolder', value: 'Attributes'}
+    - {key: 'HKEY_USERS\%%users.sid%%\Software\Classes\Wow6432Node\CLSID\*\ShellFolder', value: 'Attributes'}
+    - {key: 'HKEY_LOCAL_MACHINE\Software\Classes\CLSID\*\ShellFolder', value: 'HideOnDesktop'}
+    - {key: 'HKEY_USERS\%%users.sid%%\Software\Classes\CLSID\*\ShellFolder', value: 'HideOnDesktop'}
+    - {key: 'HKEY_LOCAL_MACHINE\Software\Classes\Wow6432Node\CLSID\*\ShellFolder', value: 'HideOnDesktop'}
+    - {key: 'HKEY_USERS\%%users.sid%%\Software\Classes\Wow6432Node\CLSID\*\ShellFolder', value: 'HideOnDesktop'}
+    - {key: 'HKEY_LOCAL_MACHINE\Software\Classes\CLSID\*\InprocServer32', value: 'LoadWithoutCOM'}
```
joachimmetz: Some of these keys already exist elsewhere, such as WindowsCOMInprocServers, please combine keys where this is appropriate

data/windows.yaml
```diff
@@ -1807,6 +1894,41 @@ urls:
...
+
+  Modifications are most easily performed via bcdedit.exe. The
+  mappings of registry key to associated bcdedit commands is as
+  follows:
+  * 16000009: 'bcdedit.exe /set {default} recoveryenabled <yes|no>'
+    * 00 gets stored for 'no', 01 gets stored for 'yes'
+  * 250000e0: 'bcdedit.exe /set {default} bootstatuspolicy ignoreallfailures'
+    * 01 00 00 00 00 00 00 00 gets stored.  Otherwise, the key is not present
+
+  The wildcard component of the registry key is the identifier associated
```
joachimmetz: registry => Windows Registry

To disambiguate

data/windows.yaml
```diff
@@ -2102,6 +2235,23 @@ urls:
 - 'https://www.ghacks.net/2016/01/08/disableosupgrade-prevents-the-upgrade-to-windows-10/'
 - 'https://blog.talosintelligence.com/2019/05/threat-roundup-0517-0524.html'
 ---
+name: WindowsUpdateSettings
+doc: |
+  Windows Update Settings
+
+  Malware can disable automatic updates in order to better avoid detection.
```
joachimmetz: > Malware can disable automatic updates in order to better avoid detection.

I opt to remove this. It does not help describe the artifact

data/windows.yaml
```diff
@@ -1807,6 +1894,41 @@ urls:
...
+doc: |
+  Windows Boot Configuration Settings.
+
+  These registry keys are associated with the Windows Boot
+  Configuration Data. Malware can modify these keys to disable
+  recovery options like the ability to boot into safe mode.
+  One malware family known to modify these keys is Cerber, a
+  well-known ransomware.
+
+  Modifications are most easily performed via bcdedit.exe. The
```
joachimmetz: Modifications are most easily performed via bcdedit.exe. => 

bcdedit.exe can be used to modify the Windows Boot Configuration Settings.

data/windows.yaml
```diff
@@ -1807,6 +1894,41 @@ urls:
 - 'https://blog.talosintelligence.com/2019/05/threat-roundup-0517-0524.html'
 - 'https://blogs.technet.microsoft.com/platforms_lync_cloud/2017/05/05/disabling-windows-10-action-center-notifications/'
 ---
+name: WindowsBootConfigurationSettings
+doc: |
+  Windows Boot Configuration Settings.
+
+  These registry keys are associated with the Windows Boot
```
joachimmetz: registry keys => Windows Registry values ?

data/windows.yaml
```diff
@@ -1807,6 +1894,41 @@ urls:
...
+  Configuration Data. Malware can modify these keys to disable
```
joachimmetz: consistency nit:
Data => Setting

Maybe less cryptic:

Malware, like Cerber (ransomware), is known to change Windows Boot Configuration Settings and  disable recovery options like the ability to boot into safe mode.

joachimmetz:
@recvfrom thx for the updates, I have some textual nits, and request to restructure some of the definitions a bit.

data/windows.yaml
recvfrom: @joachimmetz yeah, InstalledSDB\*\DatabasePath stores the path to the SDB, and Custom\* stores the mapping between the application that the SDB identifier and the file name that it applies to.  How about renaming it to something like `WindowsApplicationShims'?  I originally added "RegistryKeys" to differentiate it from the artifact that pertained to the corresponding files on disk, but changing that artifact name to 'WindowsApplicationShimDatabaseFiles' solves that problem.

joachimmetz: Maybe split them in:
* WindowsApplicationCompatibilityInstalledShimDatabases
* WindowsApplicationCompatibilityShimDatabaseMappings

You can create additional artifact group if it is convenient to collect them at the same time

data/windows.yaml
recvfrom: The InprocServer32 keys could be moved under WindowsCOMInprocServers, but I'm not sure whether it'd make sense to move the ShellFolder keys under there (since InprocServer32 and ShellFolder are at the same hierarchy, I wonder if you can have a GUID where ShellFolder exists alongside InProcHandler32 or LocalServer32, for instance).  I figured it'd be better to keep them all together in a new artifact, since they are used together by the one technique.  Do you think it'd be better to split them up, though, keeping the ShellFolder ones as their own key?

data/windows.yaml
recvfrom: Do you have any recommendations for the name of that artifact group?

data/windows.yaml
joachimmetz: If they belong together let's keep it as as-is for now. We can always change them later. The guideline / rule of thumb is roughly to keep a definition specific to a single piece of information / data, and combine them with an artifact group if there is a logical relationship between the definitions.

data/windows.yaml
joachimmetz: Maybe `WindowsApplicationCompatibilityShims`

PR #383, opened 2020-02-18
Added Redis configuration file artifact

Onager:


data/config_files.yaml
```diff
@@ -21,8 +21,8 @@ sources:
 - type: FILE
   attributes:
     paths:
-    - '%%environ_programfiles%%\Redis\conf\redis.windows.conf'
-    - '%%environ_programfiles%%\Redis\conf\redis.conf'
+    - '%%environ_programfiles%%/Redis/conf/redis.windows.conf'
+    - '%%environ_programfiles%%/Redis/conf/redis.conf'
```
joachimmetz: please use `\` and add `separator: \` to keep it closest to the OS path as possible

PR #384, opened 2020-04-02
Added System Secrets Service (SSSD) Database File artifact definition

Karneades:


joachimmetz:
@Karneades I've renamed the artifact since a "Kerberos Ticket" is something that is part of the Kerberos protocol (see: https://en.wikipedia.org/wiki/Kerberos_(protocol)) which is not the same as the sssd database file.

joachimmetz:
@Karneades thanks for the addition.

PR #385, opened 2020-04-03
Add Windows Subsystem for Linux Profile Files

Karneades:


joachimmetz:
@Karneades why not make this part of UsersShellConfigs (https://github.com/ForensicArtifacts/artifacts/blob/d5bf9a91e4f34d3869b481cb5e46da6c2773d26b/data/unix_common.yaml#L202 ) ?

Karneades:
Should we place it in the file on which host system the artifact is found (on a Windows system) or where the target artifact is stored (Linux installation)... I'm unsure. The same goes with Git for Windows (bash etc.).

We must add the label Windows to the UserShellConfigs though and I'm unsure whether people whould search in the Linux or in the Windows file when searching these paths. Outside of GRR where the label is used.

Karneades:
I didn't find the artifacts for Git for Windows, though we could add these too, same situation there. And the same paths with the current shell configs can be used.

Karneades:
Move the part to unix common and therefore we cover now Git for Windows and WSL.

joachimmetz:
@Karneades thx, trying to have a look later today.

joachimmetz:
@Karneades I've made some changes to split the WSL paths into a Windows specific source types.

PR #393, opened 2020-06-12
Update Safari cache and add Safari Extensions to webbrowser.yaml

jipegit:
Add Safari cache WAL
Write-Ahead Logging file added
Safari cache case fixed

Add SafariExtensions


joachimmetz:
@jipegit thx for the additions

PR #402, opened 2020-10-01
Chrome storage artifacts

obsidianforensics:
Adds ChromeCookies, ChromeLocalStorage, ChromeSessionStorage, ChromeFileSystem, and ChromeIndexedDB artifacts, with the latter four also comprising a new ChromeStorage artifact group.


joachimmetz:
@obsidianforensics thx for adding, LGTM

PR #409, opened 2021-01-07
Add some files related to Microsoft Edge based on Chromium

ant1:


data/webbrowser.yaml
```diff
@@ -78,6 +78,8 @@ sources:
     - '%%users.localappdata%%\Chromium\User Data\*\Cache\*'
     - '%%users.localappdata%%\Chromium\User Data\*\Media Cache\*'
     - '%%users.localappdata%%\Chromium\User Data\*\GPUCache\*'
+    - '%%users.localappdata%%\Microsoft\Edge\User Data\*\Cache\*'
```
joachimmetz, line 81: I'm wondering if it is more logical from a usage perspective to put these under a Microsoft Edge definition instead?

joachimmetz: this can be solved with an alias

PR #429, opened 2021-06-24
Add PortProxy artifact to windows.yaml

Karneades:


data/windows.yaml
```diff
@@ -1563,6 +1563,19 @@ urls:
 - 'https://support.microsoft.com/en-us/kb/323525'
 - 'https://blog.malwarebytes.com/detections/pum-optional-disallowrun/'
 ---
+name: WindowsPortProxyConfiguration
+doc: Windows PortProxy registry keys (set by netsh portproxy command or manually).
+sources:
+- type: REGISTRY_KEY
+  attributes:
+    keys:
+    - 'HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\PortProxy\v4tov4\tcp\*'
```
joachimmetz, line 1572: Would it make sense to add support for `v4tov6`, `v6tov4`, and `v6tov6` sub keys as well ? And maybe different protocols such as `udp` that are suported by `netsh`?

joachimmetz:
@Karneades thx for the proposed changes, I'm wondering if we can/should make this definition more generic.

data/windows.yaml
Karneades: Yes, think that too to make the defintion more generic. No other keys/values are set there expect these port forwardings on my test machine.

Karneades:
Using ** results in having an entry also for the keys themselves. I use no the more specific version with normal wildcards.

Karneades:
Are you planning to merge the artifact?

joachimmetz:
I'll have another look when time permits. Thx for the reminder it likely fell through the crack since the PR was not marked as to review.

joachimmetz (approved):
LGTM

PR #430, opened 2021-06-25
Add note regarding WSL to UsersShellConfigs

Karneades:
Add note for WSL because otherwise it can't be found. Filtered in GRR for "WSL", "Linux", "Subsystem", ... found nothing. At least a note for WSL must be there. 

I was searching again for WSL, didn't find anything in the artifacts and then searched through the Github repo for WSL. Found my own contribution (haha). We already worte in the PR about if we should make a note about WSL in the artifact... today, I run into that again. Was already making a new one...

Also, inside GRR I looked first only for Windows artifacts, then switched to all. As I wrote in the PR https://github.com/ForensicArtifacts/artifacts/pull/385 I would move that to Windows because I search there for the WSL part and not in the Unix part.

### DIFF
--- a/data/unix_common.yaml
+++ b/data/unix_common.yaml
@@ -200,7 +200,7 @@ labels: [Authentication]
 supported_os: [Linux, Darwin]
 ---
 name: UsersShellConfigs
-doc: Common Unix user shell configuration files.
+doc: Common Unix user shell configuration files on *nix or Windows Subsystem for Linux (WSL).
 sources:
 - type: FILE
   attributes:


joachimmetz:
@Karneades thx for the suggested changes. Must have missed your comments on https://github.com/ForensicArtifacts/artifacts/pull/385.

Could a WSL label help here? Though it does not look like GRR allows to filter on labels.

Adding WSL to `supported_os` might be more involved for GRR

Karneades:
I currently don't know how to best make it "foundable" in GRR and be enough generic for the repo. supported_os is incorrect for me because it's not an own OS. Add a label seems a good way.

joachimmetz:
@Karneades the definitions now support aliases (GRR not yet) but that might be a more suitable solution for the use case you describe.

Karneades:
timeout. closing.

PR #454, opened 2022-02-14
New artifacts

binglot:
Multiple new artifacts and updates to the existing artifacts.


data/applications.yaml
```diff
@@ -92,3 +92,16 @@ supported_os: [Windows]
...
+sources:
+- type: FILE
+  attributes:
+    paths:
+    - '%%environ_programfiles%%\Microsoft SQL Server\*\MSSQL\LOG\ERRORLOG*'
+    - '%%environ_programfilesx86%%\Microsoft SQL Server\*\MSSQL\LOG\ERRORLOG*'
+    separator: '\'
+labels: [Software, Logs]
+supported_os: [Windows]
+urls: ['https://github.com/EricZimmerman/KapeFiles/blob/a60491344a6c24ad1e6987f66283da960b9be739/Targets/Logs/MSSQLErrorLog.tkape']
```
joachimmetz, line 107: Please link to a source that is more informative about the artifact not some other definition format

joachimmetz: maybe create a forensics wiki article for this artifact? or leave the link out? seems to be fairly self explanatory the currently linked article adds no value

data/webbrowser.yaml
```diff
@@ -142,6 +142,207 @@ supported_os: [Windows,Darwin,Linux]
...
+    - '%%users.homedir%%/snap/chromium/common/chromium/*/GPUCache/*'
+    - '%%users.homedir%%/snap/chromium/common/chromium/*/Media Cache/*'
+    - '%%users.homedir%%/snap/chromium/common/chromium/Cache/*'
+    - '%%users.homedir%%/snap/chromium/common/chromium/PnaclTranslationCache/*'
+  supported_os: [Linux]
+supported_os: [Windows,Darwin,Linux]
+labels: [Browser]
+urls:
+- 'https://artifacts-kb.readthedocs.io/en/latest/sources/webbrowser/ChromeCache.html'
+- 'https://forensicswiki.xyz/wiki/index.php?title=Google_Chrome#Chrome-derived_Browsers'
```
joachimmetz, line 344: this is already linked from the previous article, I opt to remove it

data/antivirus.yaml
```diff
@@ -26,8 +32,12 @@ sources:
 - type: FILE
   attributes:
     paths:
-    - '%%environ_allusersappdata%%\Microsoft\Windows Defender\Support\MPLog-*.log'
+    - '%%environ_allusersappdata%%\Microsoft\Windows AntiMalware\Support\MPDetection-*.log'
+    - '%%environ_allusersappdata%%\Microsoft\Windows AntiMalware\Support\MPLog-*.log'
+    - '%%environ_allusersappdata%%\Microsoft\Windows Defender\Scans\History\Service\DetectionHistory\**'
     - '%%environ_allusersappdata%%\Microsoft\Windows Defender\Support\MPDetection-*.log'
+    - '%%environ_allusersappdata%%\Microsoft\Windows Defender\Support\MPLog-*.log'
+    - '%%environ_systemroot%%\Temp\MpCmdRun.log'
```
joachimmetz, line 40: is this correct? or should this be `%%users.temp%%` ?

data/antivirus.yaml
```diff
@@ -105,18 +115,23 @@ sources:
...
 name: SymantecAVQuarantine
-doc: Symantec Anti-Virus Quarantine (Infected) files.
+doc: Symantec Anti-Virus quarantine (infected) and cloud submission files.
 sources:
 - type: FILE
   attributes:
-    paths: ['%%environ_allusersappdata%%\Symantec\Symantec Endpoint Protection\**5\*.vbn']
+    paths:
+    - '%%environ_allusersappdata%%\Symantec\Symantec Endpoint Protection\**5\*.vbn'
+    - '%%environ_allusersappdata%%\Symantec\Symantec Endpoint Protection\Quarantine\**'
```
joachimmetz, line 132: Isn't the directory structure `*\*-*-*-*` instead of broad `**` ?

data/windows.yaml
```diff
@@ -3447,3 +3483,37 @@ sources:
...
+  UAL is a local data aggregation feature (enabled by default) on Windows
+  Servers 2012 and above, recording client usage by role and product on each
+  system providing the resource. It's typically between 2 and 4 extensible
+  storage engine (ESE) databases ("Current.mdb", "SystemIdentity.mdb, and
+  "<GUID>.mdb").
+sources:
+- type: FILE
+  attributes:
+    paths:
+    - '%%environ_systemroot%%\System32\LogFiles\SUM\*.mdb'
```
joachimmetz, line 3515: since this is a single path definition use the form `paths: ['...']`

data/windows.yaml
```diff
@@ -320,6 +320,30 @@ urls:
 - 'https://www.fireeye.com/content/dam/fireeye-www/global/en/current-threats/pdfs/wp-windows-management-instrumentation.pdf'
 - 'https://github.com/libyal/dtformats/blob/main/documentation/WMI%20repository%20file%20format.asciidoc'
 ---
+name: WindowsCIMDatabase
```
joachimmetz, line 323: There is already WindowsCIMRepositoryFiles, btw this definition is also not correct. The CIM repository consists of more than OBJECTS.DATA

data/windows.yaml
```diff
@@ -320,6 +320,30 @@ urls:
...
+  attributes:
+    paths:
+    - '%%environ_systemroot%%\System32\wbem\Repository\OBJECTS.DATA'
+    - '%%environ_systemroot%%\System32\wbem\Repository\FS\OBJECTS.DATA'
+    separator: '\'
+supported_os: [Windows]
+urls:
+- 'https://github.com/libyal/dtformats/blob/main/documentation/WMI%20repository%20file%20format.asciidoc'
+- 'https://www.fireeye.com/blog/threat-research/2016/12/do_you_see_what_icc.html'
+- 'https://github.com/davidpany/WMI_Forensics'
```
joachimmetz, line 345: Maybe add this link to the forensics wiki article about WMI https://forensicswiki.xyz/wiki/index.php?title=WMI

joachimmetz:
Left some initial comments but maybe break this up in multiple CL, might be easier to check/discuss some of the changes.

data/antivirus.yaml
joachimmetz: never mind I might be confusing this with Defender

if you have a documentation link that could help

data/webbrowser.yaml
```diff
@@ -19,7 +19,7 @@ sources:
 - type: ARTIFACT_GROUP
   attributes:
     names:
-    - 'ChromeHistory'
+    - 'ChromeAndDerivedBrowsersHistory'
```
joachimmetz, line 22: why remove ChromeHistory ?

data/webbrowser.yaml
```diff
@@ -142,6 +142,207 @@ supported_os: [Windows,Darwin,Linux]
 labels: [Browser]
 urls: ['https://artifacts-kb.readthedocs.io/en/latest/sources/webbrowser/ChromeCache.html']
 ---
+name: ChromeAndDerivedBrowsersCache
```
joachimmetz, line 145: Isn't Chromium the base https://en.wikipedia.org/wiki/Chromium_(web_browser)

ChromeAndDerivedBrowsersCache => ChromiumBasedBrowsersCache

data/antivirus.yaml
binglot: According to the source, it was C:\Windows\Temp, and not %%users.temp%%.

data/antivirus.yaml
binglot: I only have the path.

data/applications.yaml
binglot: Good point, I'll leave the link out. It was mainly to give the credits to the people who I copied the information from. Thanks :)

data/webbrowser.yaml
binglot: The new artifact "ChromeAndDerivedBrowsersHistory" already includes Chrome, so it's more of replacing the "ChromeHistory" artifact with a more complete "Chrome History + derived browsers".

data/webbrowser.yaml
binglot: Good catch, I didn't know the distinction between Chrome and Chromium. I'll update the descriptions so it doesn't say Chrome-based but rather Chromium-based browsers.

data/webbrowser.yaml
binglot: Sure, done.

data/windows.yaml
binglot: I realise that it's missing the other files, but that was the intent of this definition. I only wanted to grab OBJECTS.DATA, which contains the data itself (without the definitions on how to parse it, etc.). AFAIK, if you run scripts that carve out various data structures as linked in the 'urls' section, then this file is all you need, and collecting others is unnecessary (and reduces the size of what we collect). WDYT?

data/windows.yaml
binglot: Awesome, done! :)

data/windows.yaml
binglot: My oversight, fixed! 👍

data/applications.yaml
joachimmetz: then I opt to create a forensics wiki article, typically the credit does not need go to a single person or project, eg the file you link to seems to be authored by someone else than the project maintainer, but information about the error log is something that likely originates from Microsoft https://docs.microsoft.com/en-us/sql/relational-databases/performance/view-the-sql-server-error-log-sql-server-management-studio?view=sql-server-ver15

data/antivirus.yaml
joachimmetz: Are you sure it is not using %Temp% and C:\Windows\Temp was just a single observation of this ?

data/windows.yaml
joachimmetz: I would just collect all the files, since there are many caveats to reconstructing data from OBJECTS.DATA files

data/antivirus.yaml
binglot: I just did some research and it appears that in older versions of Windows [1][2] it is indeed C:\Windows\Temp and more recently it became %Temp%. It's also possible that the file would reside in C:\Windows\ServiceProfiles\LocalService\AppData\Local\Temp [3]. I'll add those to the definition to cover all 3 variants.

[1] https://www.tenforums.com/general-support/165610-location-mpcmdrun-log.html
[2] https://www.tenforums.com/general-support/165610-location-mpcmdrun-log.html

data/windows.yaml
binglot: As per our side chat, I'll remove this definition and thanks for your work on https://github.com/libyal/dtformats/blob/main/scripts/wmi_repository.py 🥇

binglot: I've updated the URLs in the definition and added the paths of the "copies created during system rebuild".

joachimmetz (approved):
LGTM, made some additional small changes.

data/webbrowser.yaml
joachimmetz: ah so ChromeHistory can likely be moved to legacy then https://github.com/ForensicArtifacts/artifacts/issues/455

PR #458, opened 2022-02-16
Label updates

binglot:
Expanded the labels, changed 'Antivirus' to 'Security Agents', and made some minor adjustments - added space between label names (e.g. 'ExternalAccount' became 'External Account') to make them consistent with others (e.g. 'History Files'), and updated the list of labels in the docs.

### DIFF
--- a/artifacts/definitions.py
+++ b/artifacts/definitions.py
@@ -13,7 +13,9 @@ TYPE_INDICATOR_WINDOWS_REGISTRY_VALUE = 'REGISTRY_VALUE'
 TYPE_INDICATOR_WMI_QUERY = 'WMI'
 
 LABELS = {
-    'Antivirus': 'Antivirus related artifacts, e.g. quarantine files.',
+    'Applications': (
+        'Configuration files and logs of installed applications '
+        'e.g. Nginx log files.'),
     'Authentication': 'Authentication artifacts.',
     'Browser': 'Web Browser artifacts.',
     'Cloud': 'Cloud applications artifacts.',
@@ -22,24 +24,30 @@ LABELS = {
     'Containerd': 'Containerd artifacts',
     'Docker': 'Docker artifacts.',
     'Execution': 'Contain execution events.',
-    'ExternalAccount': (
+    'External Account': (
         'Information about any user accounts e.g. username, '
         'account ID, etc.'),
     'External Media': 'Contain external media data or events e.g. USB drives.',
+    'File System': 'Artifacts related to the file system e.g. MFT.',
     'Hadoop': 'Hadoop artifacts.',
     'IM': 'Instant Messaging / Chat applications artifacts.',
+    'Interactive': 'Artifacts created by interactive (e.g. RDP) user activity.',
     'iOS': 'Artifacts related to iOS devices connected to the system.',
     'History Files': 'History files artifacts e.g. .bash_history.',
-    'KnowledgeBase': 'Artifacts used in knowledge base generation.',
+    'Knowledge Base': 'Artifacts used in knowledge base generation.',
     'Kubernetes': 'Kubernetes artifacts',
     'Logs': 'Contain log files.',
     'Mail': 'Mail client applications artifacts.',
     'Memory': 'Artifacts retrieved from memory.',
     'Network': 'Describe networking state.',
+    'Persistence' : 'Persistence mechanisms e.g. the Startup folder.',
     'Plist': 'Artifact that is a plist.',
     'Processes': 'Describe running processes.',
     'Rekall': 'Artifacts using the Rekall memory forensics framework.',
-    'Software': 'Installed software.',
+    'Registry': 'Files related to the Windows Registry.',
+    'Security Agents': 'Endpoint detection and response related artifacts, '
+        'e.g. antivirus quarantine files.',
+    'Software': 'Information about installed software.',
     'SQLiteDB': 'Artifact that is a SQLite database.',
     'System': 'Core system artifacts.',
     'Users': 'Information about users.'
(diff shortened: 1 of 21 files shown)


data/macos.yaml
```diff
@@ -314,7 +314,7 @@ urls:
 - 'https://forensicswiki.xyz/wiki/index.php?title=Mac_OS_X'
 - 'https://forensicswiki.xyz/wiki/index.php?title=Mac_OS_X_10.9_-_Artifacts_Location#Kernel_Extension'
 ---
-name: MacOSDuetKnowledgeBase
+name: MacOSDuetKnowledge Base
```
joachimmetz, line 317: Please change this back, looks like one too many find replace ;)

binglot: Haha, my `sed` command was too effective ;)

artifacts/definitions.py
```diff
@@ -22,24 +25,32 @@
...
-    'ExternalAccount': (
+    'ExternalAccount': '<Deprecated>', # TODO: to be deleted
+    'External Account': (
         'Information about any user accounts e.g. username, '
         'account ID, etc.'),
     'External Media': 'Contain external media data or events e.g. USB drives.',
+    'File System': 'Artifacts related to the file system e.g. MFT.',
     'Hadoop': 'Hadoop artifacts.',
     'IM': 'Instant Messaging / Chat applications artifacts.',
+    'Interactive': 'Artifacts created by interactive (e.g. RDP) user activity.',
```
joachimmetz, line 36: This label might be misleading, since a lot of "interactive" artifacts can be script as well.

artifacts/definitions.py
```diff
@@ -22,24 +25,32 @@
...
     'History Files': 'History files artifacts e.g. .bash_history.',
-    'KnowledgeBase': 'Artifacts used in knowledge base generation.',
+    'KnowledgeBase': '<Deprecated>', # TODO: to be deleted
+    'Knowledge Base': 'Artifacts used in knowledge base generation.',
     'Kubernetes': 'Kubernetes artifacts',
     'Logs': 'Contain log files.',
     'Mail': 'Mail client applications artifacts.',
     'Memory': 'Artifacts retrieved from memory.',
     'Network': 'Describe networking state.',
+    'Persistence' : 'Persistence mechanisms e.g. the Startup folder.',
```
joachimmetz, line 46: This label likely looses its value since there are many persistence mechanisms

artifacts/definitions.py
```diff
@@ -22,24 +25,32 @@
...
     'Plist': 'Artifact that is a plist.',
     'Processes': 'Describe running processes.',
     'Rekall': 'Artifacts using the Rekall memory forensics framework.',
-    'Software': 'Installed software.',
+    'Registry': 'Files related to the Windows Registry.',
```
joachimmetz, line 50: What is the purpose of this label?

artifacts/definitions.py
```diff
@@ -22,24 +25,32 @@
...
+    'Security Agents': 'Endpoint detection and response related artifacts, '
+        'e.g. antivirus quarantine files.',
+    'Software': 'Information about installed software.',
```
joachimmetz, line 53: How is this label different from Applications ?

joachimmetz:
I opt to split the deprecation of labels and standardizing on having a space, versus adding new labels, and renaming labels

joachimmetz (changes requested):
left some questions, the purpose of some of the proposed labels is unclear to me

artifacts/definitions.py
joachimmetz: Also there are already groups for this purpose such WindowsPersistenceMechanisms and WindowsPersistenceRegistryKeys. Why is an additional label needed? what is the use case?

artifacts/definitions.py
binglot: This label refers to artifacts that would be generated through GUI interaction with the system (e.g. ShellBags in Windows), which to you point could be both automated and manual. What do you suggest -- make the comment more clear, rename the label to something different?

artifacts/definitions.py
binglot: The use case is to iterate over all artifact definitions and pick up those that have the label Persistence. What's the downside of having the label and adding it to all artifact definitions that relate to maintaining persistence?

artifacts/definitions.py
binglot: I admit that maybe there's not much value in having a label for the registry files (hives, transaction files, backups, etc.). It just felt right to also assign them a label :) but I'm also happy to delete this label. WDYT?

artifacts/definitions.py
binglot: The distinction I was trying to make is that "Applications" relates to artifacts related directly to software, such as logs and configuration files, while "Software" would relate to what software was installed / software inventory (e.g. Python packages, Chrome extensions, the Uninstall key in the Windows registry, etc.). Maybe there's a better way to draw that distinction by renaming "Software" to "Software Inventory", WDYT?

artifacts/definitions.py
joachimmetz: If we're not using it remove it

artifacts/definitions.py
joachimmetz: How do you want to use this label?

artifacts/definitions.py
joachimmetz: > What's the downside of having the label and adding it to all artifact definitions that relate to maintaining persistence?

note that this label is applicable to a lot of definitions. If it is added to most of the artifact definitions the label, the label looses its value as way to easily filter on persistence related artifacts.

artifacts/definitions.py
joachimmetz: But what would be the use case, I think labels are mainly used in GRR at the moment and by the stats script https://artifacts.readthedocs.io/en/latest/sources/background/Stats.html. Maybe there are other ways / categories that could be more fitting

binglot:
As discussed offline with @joachimmetz , I'll close this PR because the labels will be completely deprecated. Instead, one can use Artifact Groups.

PR #466, opened 2022-02-19
Update Dropbox definition to include sync_history.db

sydp:
Add database file procesed by plaso parser (ref: https://github.com/log2timeline/plaso/pull/4052)

### DIFF
--- a/data/cloud_services.yaml
+++ b/data/cloud_services.yaml
@@ -20,7 +20,7 @@ sources:
     paths:
     - '%%users.appdata%%\Dropbox\*.db*'
     - '%%users.localappdata%%\Dropbox\*.db*'
-    - '%%users.homedir%%/.dropbox/instance1/sync_history.db'
+    - '%%users.localappdata%%\Dropbox\instance1\sync_history.db'
     separator: '\'
   supported_os: [Windows]
 - type: FILE
--- a/data/cloud_services.yaml
+++ b/data/cloud_services.yaml
@@ -20,12 +20,14 @@ sources:
     paths:
     - '%%users.appdata%%\Dropbox\*.db*'
     - '%%users.localappdata%%\Dropbox\*.db*'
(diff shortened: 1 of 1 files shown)


joachimmetz:
@sydp FYI the validation test is failing please check the definition:

```
======================================================================
FAIL: testArtifactDefinitionsValidator (validator_test.ArtifactDefinitionsValidatorTest)
Runs the validator over all the YAML artifact definitions files.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\projects\artifacts\tests\validator_test.py", line 24, in testArtifactDefinitionsValidator
    self.assertTrue(
AssertionError: False is not true : in definitions file: data\cloud_services.yaml
```

data/cloud_services.yaml
```diff
@@ -20,12 +20,14 @@ sources:
     paths:
     - '%%users.appdata%%\Dropbox\*.db*'
     - '%%users.localappdata%%\Dropbox\*.db*'
+    - '%%users.homedir%%/.dropbox/instance1/sync_history.db'
```
joachimmetz, line 23: you'll need to add a separate source type for this

sydp: My late night Ctrl+C -V failed me.  I've updated with the correct value and local unit tests appear to work.

data/cloud_services.yaml
```diff
@@ -20,12 +20,14 @@ sources:
     paths:
     - '%%users.appdata%%\Dropbox\*.db*'
     - '%%users.localappdata%%\Dropbox\*.db*'
+    - '%%users.localappdata%%\Dropbox\instance1\sync_history.db'
```
joachimmetz, line 23: instance1 hints that there might be more instances, do you know if this can be the case?

sydp: An instance2 folder is created when you log out and log in with a different account.   

I've updated the definition to glob after ```instance```.

joachimmetz (approved):
LGTM

PR #476, opened 2022-03-02
Live Response grouping of artifact definitions

binglot:
A templated set of Artifact Groups that captures Windows artifacts and can be extended for Mac and Linux later. The purpose is to use those groups for Triage / Live Response analysis of systems.


binglot:
The tests fail because there are references to artifacts that don't exist in the repo 😞 I'll need to figure out why and fix it.

binglot:
Re-opened a new PR that's up to date with upstream - https://github.com/ForensicArtifacts/artifacts/pull/493

PR #493, opened 2022-04-19
Live Response grouping of artifact definitions

binglot:
A templated set of Artifact Groups that captures Windows artifacts and can be extended for Mac and Linux later. The purpose is to use those groups for Triage / Live Response analysis of systems.

### DIFF
--- a/data/live_response.yaml
+++ b/data/live_response.yaml
@@ -1,0 +1,172 @@
+# Live Response specific artifacts.
+---
+name: LRApplicationConfigsAndLogs
+doc: Group of configuration files and logs of installed applications.
+sources:
+- type: ARTIFACT_GROUP
+  attributes:
+    names:
+    - MicrosoftIISLogs
+    - MicrosoftSqlServerErrorLogs
+    - RedisConfigFile
+    - TomcatFiles
+    - TomcatPasswordFile
+supported_os: [Windows]
+---
+name: LRExecution
+doc: Group of process/command execution related artifacts.
+sources:
+- type: ARTIFACT_GROUP
+  attributes:
+    names:
+    - JavaCacheFiles
+#    - ListProcessesGrr
+    - WindowsAMCacheHveFile
+    - WindowsCIMRepositoryFiles
+    - WindowsCrashDumps
+    - WindowsPrefetchFiles
+    - WindowsRecentFileCacheBCF
+    - WindowsStartupInfo
+    - WindowsSuperFetchFiles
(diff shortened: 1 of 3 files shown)


binglot:
@joachimmetz - PTAL :)

data/live_response.yaml
```diff
@@ -0,0 +1,172 @@
...
+name: LRWebBrowserExtensions
+doc: Group of web browser extensions related artifacts.
+sources:
+- type: ARTIFACT_GROUP
+  attributes:
+    names:
+    - ChromiumBasedBrowsersExtensions
+    - ChromiumBasedBrowsersExtensionActivitySQLiteDatabaseFile
+    - ChromePreferences
+#    - Crease
```
joachimmetz, line 160: remove this?

binglot: Done, PTAL

data/live_response.yaml
```diff
@@ -0,0 +1,165 @@
...
+doc: Group of configuration files related artifacts.
+sources:
+- type: ARTIFACT_GROUP
+  attributes:
+    names:
+    - WindowsRegistryFilesAndTransactionLogs
+    - WindowsSystemRegistryFilesAndTransactionLogsBackup
+supported_os: [Windows]
+---
+name: LiveResponseSystemLogs
```
joachimmetz, line 135: This looks very Windows centric.

data/live_response.yaml
```diff
@@ -0,0 +1,165 @@
...
+    names:
+    - MicrosoftIISLogs
+    - MicrosoftSqlServerErrorLogs
+    - RedisConfigFile
+    - TomcatFiles
+    - TomcatPasswordFile
+supported_os: [Windows]
+---
+name: LiveResponseExecution
+doc: Group of process/command execution related artifacts.
```
joachimmetz, line 17: @binglot let's chate more about the naming. Would it be clearer to name them according to their purpose e.g. TriageWindowsProcessExecution ?

joachimmetz: Also these groups do not necessary have to be used for live response (or triage for that matter).

binglot: As discussed offline, I've changed it from 'Live Response' to 'Triage'.

data/live_response.yaml
binglot: The plan is for this to evolve to logs from other platforms.

joachimmetz (approved):
LGTM

PR #506, opened 2022-06-14
Temporarily removing Windows from UsersDirectory

binglot:
The e2e tests of GRR started to break due to the UsersDirectory supporting Windows. The issue is that this definition uses "provides: [users.username]" and the Knowledge Base used by GRR relies on it, so suddenly it has an unexpected list of users.

As a temporary solution until we figure out a better one, I'd ask that we remove the Windows support so the GRR builds can pass again.


binglot:
@joachimmetz - PTAL :)

joachimmetz (commented):
LGTM

PR #522, opened 2022-07-26
Update Linux artifacts.

sydp:
Additional Linux artifacts:

- Debian APT history term
- Systemd journal logs
- Gnome tracker files (databases)


joachimmetz (approved):
LGTM

PR #526, opened 2022-08-19
Add a script to print artifact definitions to the console.

sydp:
The purpose of this tool script is to convert (and flatten) artifact YAML definition files into a CSV/JSONL format which allows easier importing to spreadsheets (Google Sheets or MS Excel) or pandas dataframes.


tools/list_artifacts.py
```diff
@@ -0,0 +1,213 @@
...
+      directory: the data directory of artifact definitions.
+      target_os: the target operating system.
+      split_paths: split artifict definitions into separate entries when
+         there are more than one path values."""
+    self.directory = directory
+    self.target_os = target_os
+    self.split_paths = split_paths
+
+  def _LoadArtifacts(self):
+    """Generates a dictionary representing an artifact."""
```
joachimmetz, line 37: add Yields information to docstring.

Why not break this large function up in smaller functions for the different type of source types ?

tools/list_artifacts.py
```diff
@@ -0,0 +1,213 @@
...
+
+  if not options.definitions:
+    print('Source value is missing.')
+    print('')
+    args_parser.print_help()
+    print('')
+    return False
+
+  if not os.path.exists(options.definitions):
+    print('No such file or directory: {0:s}'.format(options.definitions))
```
joachimmetz, line 194: This codebase just switched to fstrings

tools/list_artifacts.py
```diff
@@ -0,0 +1,213 @@
...
+  """
+
+  def __init__(self, directory, target_os, split_paths=False):
+    """Initializes an artifact lister.
+
+    Args:
+      directory: the data directory of artifact definitions.
+      target_os: the target operating system.
+      split_paths: split artifict definitions into separate entries when
+         there are more than one path values."""
```
joachimmetz, line 31: Put closing `"""` on a new line

tools/list_artifacts.py
sydp: Done

tools/list_artifacts.py
sydp: Done

tools/list_artifacts.py
sydp: Done

joachimmetz:
@sydp what is the status of this PR? 

Could you provide me with some context what you mean with "convert (and flatten)" and why this is needed?

tools/list_artifacts.py
```diff
@@ -0,0 +1,327 @@
...
+    args_parser.print_help()
+    print('')
+    return False
+
+  if not os.path.exists(options.definitions):
+    print(f'No such file or directory: {options.definitions}')
+    print('')
+    return False
+
+  artifact_lister = ArtifactLister(
```
joachimmetz, line 312: why not use ArtifactDefinitionsRegistry ? given that you want to make sure you have the necessary definitions for groups.

joachimmetz:
@sydp be aware that scripts have moved https://github.com/ForensicArtifacts/artifacts/commit/71669eafffa22aa71a113dce27450a023fed0346

sydp:
> @sydp what is the status of this PR?
> 
> Could you provide me with some context what you mean with "convert (and flatten)" and why this is needed?
> 

It's needed to represent a path within a definition as a CSV.   It was useful for me in doing some analysis of existing artifacts (and to update/add linux-based artifacts) and could be useful for anyone else that might want to get a representation of the artifacts in this format.  If it doesn't make sense to be part of this repo, it can be closed.

joachimmetz:
So if I understand you correctly your use case here is to be able to convert the yaml to csv. Given there are tools like yaml2csv that specialize in this IMHO I see no strong need for this script. 

> If it doesn't make sense to be part of this repo, it can be closed.

At the moment I fail to see a strong reason versus the added maintenance, closing per suggestion.

PR #563, opened 2023-05-30
Modified syslog glob to match more files

Onager:
Notably on Ubuntu, the syslog file is /var/log/syslog, not /var/log/syslog.log


joachimmetz:
changed regex to glob, waiting for CI tests to complete

PR #608, opened 2024-01-30
Make WindowsRecycleBin use absolute path.

The-Briel-Deal:
Since all other artifacts here resolve to an absolute path, it would be nice if this followed the same convention and included the partition letter as well.

### DIFF
--- a/artifacts/data/windows.yaml
+++ b/artifacts/data/windows.yaml
@@ -2051,8 +2051,8 @@ sources:
 - type: FILE
   attributes:
     paths:
-    - '\$Recycle.Bin\**'
-    - '\Recycler\**'
+    - '%%environ_systemdrive%%\$Recycle.Bin\**'
+    - '%%environ_systemdrive%%\Recycler\**'
     separator: '\'
 supported_os: [Windows]
 urls: ['https://forensics.wiki/windows#recycle-bin']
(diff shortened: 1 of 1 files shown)


joachimmetz:
First of all a path starting with `\` is absolute (from the volume root)

`'%%environ_systemdrive%%` only refers to (typical) `C:` not other drive letters where the recycle bin can also be on other volumes. One alternative approach is to add an indicator to indicate that this should apply to all volumes `%%system_all_volumes%%` or equiv.

The-Briel-Deal:
> First of all a path starting with `\` is absolute (from the volume root)
> 
> `'%%environ_systemdrive%%` only refers to (typical) `C:` not other drive letters where the recycle bin can also be on other volumes. One alternative approach is to add an indicator to indicate that this should apply to all volumes `%%system_all_volumes%%` or equiv.
> 
> Also see: https://dfvfs.readthedocs.io/en/latest/sources/developer/Helpers.html#windows-path-resolver-helper

Thats a good point, we decided to put this in our backlog for now. Once we add a variable in GRR like %%system_all_volumes%% we will come back to this.

Closing for now. We'll reopen once the appropriate variable exists. Thanks for the quick response!

joachimmetz:
Created https://github.com/ForensicArtifacts/artifacts/issues/609 as a reminder

Issue #275, opened 2018-07-18
Deprecate provides


joachimmetz:
Deprecate provides

Per GRR team, provides are used in interpolation dependency resolving, but provides make GRR more fragile.

* [x] remove provides from GRR
* [x] undo changes in https://github.com/ForensicArtifacts/artifacts/pull/506 seeing this currently breaks GRR tests
  * https://github.com/ForensicArtifacts/artifacts/pull/612

joachimmetz:
Grr uses provides in SetKBValue:
https://github.com/google/grr/blob/68a45fd446704285af25403d8dde4d3630524165/grr/server/grr_response_server/artifact.py#L233

joachimmetz:
linux.yaml
```
os_release
os_major_version
os_minor_version
users.username
users.last_logon
```

macos.yaml
```
users.username
```

windows.yaml
```
users.cookies
users.appdata
users.personal
users.startup
users.homedir
users.desktop
users.internet_cache
users.localappdata
users.localappdata_low
users.recent
users.userprofile
users.temp

code_page]
current_control_set]
domain]
environ_allusersappdata]
environ_allusersprofile]
environ_path]
environ_profilesdirectory]
environ_programfiles]
environ_programfilesx86]
environ_systemdrive]
environ_systemroot]
environ_temp]
environ_windir]
time_zone
users.sid
users.userprofile
users.homedir
users.username
```

wmi.yaml
```
users.homedir
users.userdomain
```

legacy.yaml
```
code_page
current_control_set
domain
environ_allusersappdata
environ_allusersprofile
environ_path
environ_programfiles
environ_programfilesx86
environ_systemdrive
environ_systemroot
environ_temp
environ_windir
os_release
os_major_version
os_minor_version
time_zone
users.username
```

joachimmetz:
Interpolation values:

```
%%environ_allusersappdata%%
%%environ_allusersprofile%%
%%environ_programfiles%%
%%environ_programfilesx86%%
%%environ_systemdrive%%
%%environ_systemroot%%
%%environ_windir%%
%%users.appdata%%
%%users.homedir%%
%%users.localappdata%%
%%users.localappdata_low%%
%%users.sid%%
%%users.temp%%
%%users.username%%
%%users.userprofile%%
```

PR #610, opened 2024-02-08
Removed support for provides #275

panhania:
GRR (which, as far as I know, was the only blocker for the deprecation) does not rely on this feature anymore.


joachimmetz:
Thanks I'll take a look later this week

joachimmetz (approved):
LGTM

PR #619, opened 2024-03-15
Change WindowsServices to include subkeys and values

chb2mn:
HKLM\System\CurrentControlSet\Services\ is a listing of all service names and that's about it. Adding a * includes the config data


joachimmetz:
@chb2mn I recall that the intent of this artifact definition has been to get the names of the services and drivers (https://artifacts-kb.readthedocs.io/en/latest/sources/windows/ServicesAndDrivers.html) only, not the full hierarchy under them. Do you have additional context on why you want to make this change?

PR #621, opened 2024-03-22
adding W3SVC to filepath

chb2mn:
This is the default logging path for IIS on my Windows Server 2019. If it's not a typo, this should be added as a separate line.

image

### DIFF
--- a/artifacts/data/webservers.yaml
+++ b/artifacts/data/webservers.yaml
@@ -100,8 +100,9 @@ sources:
 - type: FILE
   attributes:
     paths:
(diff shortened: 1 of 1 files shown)


joachimmetz:
Let's keep the '%%environ_systemdrive%%\inetpub\logs\LogFiles\*.log' (per https://learn.microsoft.com/en-us/iis/manage/provisioning-and-managing-iis/managing-iis-log-file-storage) and add the additional path

joachimmetz (approved):

PR #626, opened 2024-05-01
added iOS and Android as supported Operating Systems

rick-slin:
I'd like to add iOS and Android as acceptable values for the supported_os field.

Thank you for your consideration.


joachimmetz:
We'll have to update the documentation as well https://github.com/ForensicArtifacts/artifacts/blob/main/docs/sources/Format-specification.md#supported-operating-system

I'll need to check with the GRR maintainers as well if this might cause issues.

joachimmetz (approved):
LGTM

joachimmetz:
Checked with GRR maintainers, no concerns from their end.

PR #638, opened 2025-07-09
Update webbrowser.yaml with ChromeBrowserHistory snapshots

wajihyassine:
See https://support.google.com/chrome/a/answer/9917429?hl=en

For major chrome updates, the last 3 revisions of a User's browsing data are kept on the machine, in case the update causes issues for the user.

This PR is to add these artifacts for collection


joachimmetz (approved):

joachimmetz:
LGTM

PR #639, opened 2025-07-22
Add extra paths for MacOS Crontab

coperni:
Older Mac versions store crontabs at /var/cron/tabs/* and /var/at/tabs/*  when a user runs crontab -e.


joachimmetz:
Thanks for the additions, taking a closer look when time permits

joachimmetz (approved):
LGTM

PR #644, opened 2025-09-12
Added Darwin as supported_os for Fish History

coperni:
Some Darwin systems use Fish shell. This PR adds support for the OS.


joachimmetz (approved):
LGTM

PR #663, opened 2026-04-22
Changes to WindowsUserShellFolders definition

panhania:
Based on the clarification in #662, I fixed the `WindowsUserShellFolders` artifact: `HKEY_USERS\%%users.sid%%\Software\Microsoft\Windows\CurrentVersion\Explorer\Shell Folders` and `HKEY_USERS\%%users.sid%%\Environment` contain no subkeys, only values, so the glob at the end of the `keys` attribute will match nothing. `HKEY_USERS\%%users.sid%%\Volatile Environment` is a bit different as it contains both values and subkeys (with values), so the glob makes sense given we also include the key itself.

### DIFF
--- a/artifacts/data/windows.yaml
+++ b/artifacts/data/windows.yaml
@@ -3327,8 +3327,9 @@ sources:
 - type: REGISTRY_KEY
   attributes:
     keys:
-    - 'HKEY_USERS\%%users.sid%%\Software\Microsoft\Windows\CurrentVersion\Explorer\Shell Folders\*'
-    - 'HKEY_USERS\%%users.sid%%\Environment\*'
+    - 'HKEY_USERS\%%users.sid%%\Software\Microsoft\Windows\CurrentVersion\Explorer\Shell Folders'
+    - 'HKEY_USERS\%%users.sid%%\Environment'
+    - 'HKEY_USERS\%%users.sid%%\Volatile Environment'
     - 'HKEY_USERS\%%users.sid%%\Volatile Environment\*'
 supported_os: [Windows]
 ---
--- a/artifacts/data/windows.yaml
+++ b/artifacts/data/windows.yaml
@@ -3328,9 +3328,10 @@ sources:
   attributes:
     keys:
     - 'HKEY_USERS\%%users.sid%%\Software\Microsoft\Windows\CurrentVersion\Explorer\Shell Folders'
+    - 'HKEY_USERS\%%users.sid%%\Software\Microsoft\Windows\CurrentVersion\Explorer\User Shell Folders'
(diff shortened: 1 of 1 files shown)


joachimmetz:
thanks for the proposed changes, taking a closer look as soon as time permits

joachimmetz:
"HKEY_USERS\%%users.sid%%\Software\Microsoft\Windows\CurrentVersion\Explorer\Shell Folders" is a key where the shell folders are string values

there is also "HKEY_USERS\%%users.sid%%\Software\Microsoft\Windows\CurrentVersion\Explorer\User Shell Folders" currently not in the definition

"HKEY_USERS\%%users.sid%%\Environment" is a key where the environment variables are string values and "HKEY_USERS\%%users.sid%%\Volatile Environment" as well

No indication that "HKEY_USERS\%%users.sid%%\Volatile Environment" has sub keys

joachimmetz:
Also observed is `HKEY_USERS%%users.sid%%\Software\Microsoft\Windows\CurrentVersion\Explorer\User Shell Folders\New` with string values (seen on XP and win2k3)

And it looks like shell folder names can be redefined between "Shell Folders" and "User Shell Folders\New"

https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-2000-server/cc962613(v=technet.10)?redirectedfrom=MSDN

order of precedence appears to be (but needs to be confirmed)

1. HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Explorer\User Shell Folders (Highest Precedence)
2. HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Explorer\User Shell Folders\New (Secondary / Legacy Fallback)
3. HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Explorer\Shell Folders (Lowest Precedence / Deprecated)
4. Similar keys in HKEY_LOCAL_MACHINE

joachimmetz:
These definitions date back to https://github.com/ForensicArtifacts/artifacts/commit/a214f24260802a36a21b3ef8a0a3c201a19e7432#diff-e96cfeffeb8496903195fbe45463f54bee773005d51fdba6bf25054cbc3c8f49R16-R79 and seem to have originated from the GRR codebase, in case more historical context is needed

joachimmetz:
@panhania note that technically the Environment keys are not user shell folders, let me know if these should/can be split into multiple definitions?

panhania:
> note that technically the Environment keys are not user shell folders, let me know if these should/can be split into multiple definitions?

I don't think it matters from the GRR perspective as we do not use artifacts for internal flow logic (like interrogation) anymore. So entirely up to your preference how these are organised.

PR #670, opened 2026-05-14
Fix: missing whitespace in unix_common.yaml

stbe:
Whitespace was missing, leading to invalid YAML for one of the FILE sources of CupsJobCacheFile


joachimmetz:
@stbe good catch, thanks for the changes

